PR #18, opened 2020-03-09
hashtable-linear: changed hashtable implementation from cuckoo to

benjaminselfridge:
linear due to bugs and lack of support for the former

Same as change for parameterized-utils

### DIFF
--- a/what4/src/What4/Expr/Builder.hs
+++ b/what4/src/What4/Expr/Builder.hs
@@ -178,7 +178,7 @@ import qualified Data.Binary.IEEE754 as IEEE754
 import qualified Data.Bits as Bits
 import           Data.Foldable
 import qualified Data.HashTable.Class as H (toList)
-import qualified Data.HashTable.ST.Cuckoo as H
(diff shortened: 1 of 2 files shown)


robdockins:
@kquick, should we retool this to use the `basic`  hashtable instead? That was my latest understanding from the performance regression tests we were doing.

kquick:
I would like to validate that by making a hashtable-basic branch for what4 and comparing the build results from those to evaluate this performance separately than the parameterized-utils change.  I will do that momentarily.

kquick:
Performance results between this branch and PR #21 shows that linear's performance is the worst, that basic is never worse than linear and can sometimes be as good as cuckoo (or  even _slightly_ better, although in general cuckoo is the most performant).  Based on these results, PR #21 should be preferred over this PR.

kquick:
Those tests also confirm that this change does impact performance of the saw verification tests, although not quite to the degree of the corresponding changes in parameterized-utils.

robdockins:
Thanks for looking into that, @kquick.  I'll close this and plan to merge #21.

PR #31, opened 2020-04-16
Bitvector abstract domain enhancements

robdockins:
The bitvector abstract domain has been split into two subdomains: one for dealing with arithmetic operations and one for dealing with bitwise operations.  We attempt to minimize information loss by only shifting between the two modes when necessary.  The Cryptol specification of the these domains is likewise split between the two modes.

I've added a new test suite specifically to test the same correctness assertions as found in the Cryptol specification.  This exposed bugs in the `sdiv` and `srem` operations, where the implementations did not meet the specification.

There is also a third domain optimized for performing XOR operations.  It is essentially isomorphic to the bitwise domain, but requires fewer operations to compute XOR.  It is specifically for use in the bitwise XOR semi-ring.

Finally, fill in a few missing operations: there are now abstract implementations for rotates, popcount, count leading zeros and count trailing zeros.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,13 +47,6 @@ matrix:
   allow_failures:
   - env: CABALVER=2.4 GHCVER=8.6.3 BUILD_ARG= DO_LINT=True
 
-cache:
-  directories:
-  - $HOME/.cabsnap
-  - $HOME/.cabal/packages
-  - $HOME/.cabal/store
-  - $HOME/.ghc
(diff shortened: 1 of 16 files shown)


robdockins:
This is rebased onto recent master, and I'm done with all the things I had planned. I think this is good to go.

robdockins:
@brianhuffman, I know you had a quick look at this, any comments before I merge?

PR #37, opened 2020-05-08
Expr builder reorg

robdockins:
This is a relatively large-scale reorganization of some of the aspects of What4.  In particular, it allows the interpreted floating-point mode to be fully decoupled from the ExprBuilder.  Additionally, the type of "locations" that are recorded in expressions can be parameterized.  This allows the "FunctionName" and "ProgramLoc" modules to be moved back into the `crucible` package where they make more sense.

This reorganization touches a lot of code, but I hope it will have have fairly small impact on most use sites.  However, setting up and instantiating an ExprBuilder now requires a bit of additional work (instantiating some type families) and the usage of some of the expr builder type indices has changed.


robdockins:
This refactoring is not ready yet, so I'm closing this for now.

PR #43, opened 2020-06-05
Make ADT field comments normal, not haddock for internal structure.

kquick:
GHC 8.4 haddock does not support comments on individual fields of an
ADT.  This particular structure is internal only and lacks general
description haddocks, so simply converting these from haddock comments
to normal comments resolves this issue.


benjaminselfridge:
Fine with me.

Issue #44, opened 2020-06-10
Proposed added exports to SMTWriter


Ptival:
For some of my purposes in building a Sally programs from Crucible programs, I am trying to use What4 as a glorified pretty-printer (with minimal solving involved). Unfortunately, the Sally syntax requires me to do non-SMTLib2-compliant operations to the produced expressions.

Given the current infrastructure of SMTWriter, I got to a working solution by exposing the following symbols:
```
CollectorResults(..)
mkBaseExpr
runInSandbox
typeMap
```

The first three let me build my own writer, and intercept the collector results, as I process them in a ad hoc way.

The latter I'm using to get from a `What4.BaseTypeRepr` to a `SMT2.Sort`, and am wondering whether I am missing a helper to do so.

Would these be okay to expose?
Eventually, I have larger ideas on how to reformulate `SMTWriter` to allow for the flexibility I need without "leaking" the aforementioned functions, but in the meantime, this is currently working well for me.

PR #45, opened 2020-06-11
Expose some SMT collector operations in SMTWriter.

Ptival:
As discussed in #44 , those changes will allow implementing language-sally by leveraging What4's SMTWriter.

Speaking of which, the language-sally package will need to depend on a version of what4 that exposes those functions. Does this warrant a new minor version number? Should I just hardcode this commit number until what4 is ready for a version bump?

EDIT: Come to think about it, I will be using language-sally within crucible and can just use a submodule to pinpoint the commit for the time being.


robdockins:
Yeah, probably just use a submodule pin.  When we start using Hackage releases of what4 this will be a more pressing problem, but for now submodules are fine.

PR #54, opened 2020-07-06
Adds testing abstraction layer to remove direct test library dependency.

kquick:


what4/what4.cabal
```diff
@@ -204,16 +205,36 @@ test-suite bvdomain_tests
...
                , parameterized-utils
                , tasty >= 0.10
                , tasty-quickcheck >= 0.10
                , QuickCheck
+               , transformers
+               , what4
+
+  ghc-options: -Wall -Wcompat -Werror=incomplete-patterns -Werror=missing-methods -Werror=overlapping-patterns
+
+test-suite bvdomain_tests_hh
```
robdockins, line 223: Having two `test-suite` stanzas will run the test both ways right? Maybe we should use a cabal flag to choose between them instead.

what4/what4.cabal
```diff
@@ -204,16 +205,36 @@ test-suite bvdomain_tests
...
   main-is: BVDomTests.hs
 
-  other-modules:
+  other-modules:  VerifyBindings
 
   build-depends: base
                , parameterized-utils
                , tasty >= 0.10
                , tasty-quickcheck >= 0.10
                , QuickCheck
```
robdockins, line 217: We should do a little bit of investigating to find some reasonable package bounds for QC.  As I recall QC version 2 made a lot of breaking changes. I don't remember when the `Gen` monad was introduced, for example.

what4/src/Test/Verification.hs
```diff
@@ -0,0 +1,194 @@
...
+
+-- | This is the generator monad for the Verification proxy tests.
+-- The inner monad will be the actual test implementation's monadic
+-- generator, and the 'a' return type is the type returned by running
+-- this monad.
+--
+-- Tests should only use the 'GenV m TYPE' as an output; the
+-- constructors and internals should be used only by the test
+-- concretization.
+newtype GenV m a = GenV { unGenV :: ReaderT (GenEnv m) m a }
```
robdockins, line 158: Can we go even further?  What about:
```
data GenV a = GenV (forall m. Monad m => ReaderT (GenEnv m) m a)
```

what4/src/Test/Verification.hs
```diff
@@ -0,0 +1,194 @@
...
+  deriving (Applicative, Functor, Monad)
+
+
+-- | A test generator that returns True or False
+chooseBool :: Monad m => GenV m Bool
+chooseBool = do gs <- GenV $ asks genChooseBool
```
robdockins, line 164: I think you can do `GenV (asks genChooseBool >>= lift)`

what4/what4.cabal
kquick: It generates two executables: one that uses QuickCheck and one that uses Hedgehog.  A cabal flag could choose but then we'd only build one of those: this way I'm actually verifying that the QuickCheck and Hedgehog mapping both work.

what4/what4.cabal
kquick: Good point.  We can probably re-introduce the original bounds and then see if they could be loosened or if they are what we want.

what4/src/Test/Verification.hs
kquick: Interesting: that would reduce the impact on test sites.  Renaming the type from `GenV` back to `Gen` would make it almost a no-change drop-in replacement.

kquick:
Regarding a separate module for test writers v.s. implementation mapping: maybe.  It seemed like pushing too hard in that direction might get really complicated and I wasn't even sure it could be teased apart like that, so I didn't try it.

Longer term/separate package:  maybe.  I think this type of abstraction could be useful for other libraries as well, but I worry that it's really only valid for a fairly constrained subset of test code.  The what4 tests fit that constraint, but the more complicated the test generation and testing code gets, the more it might need to use testing functionality that isn't really common.  I figured this was something we could do down the road if we wanted to though without too much impact.

what4/src/Test/Verification.hs
```diff
@@ -0,0 +1,194 @@
...
+-- QuickCheck Property or a Hedgehog Property.  The 'toNativeProperty'
+-- implementation function converts from these proxy Properties to the
+-- native Property implementation.
+--
+-- Tests should only use the 'Property' type as an output; the
+-- constructors and internals should be used only by the test
+-- concretization.
+data Property = BoolProperty Bool
+              | AssumptionProp Assumption
+              | PropProperty Property
```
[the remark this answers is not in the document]
kquick: I think `PropProperty` is needed for when the test code generates a property (which would be a property in this module) that needs to be "stored" for the implementation side to convert into a native `QuickCheck.Property` or `Hedgehog.Property`.  That said, elimination simplification would be nice.

what4/what4.cabal
robdockins: Turns out `chooseInt` and `chooseInteger` were both added in 2.14; they are simply higher-performance variants of the generic `choose` method.  We instead just use `choose` loosen the bounds, I think.

kquick: I was just looking at that as well.  The older `choose` is less performant, but I don't think we are inordinately slow here, so broadening the bounds is probably a better benefit.

robdockins:
I'm pretty happy with this now, what do you think @kquick?

kquick:
I've confirmed a lower-bound of 2.12 for QuickCheck still builds and runs.  I'm happy with it, too and will merge it shortly.

PR #55, opened 2020-07-08
App refactor

robdockins:
This pulls in the parts of `expr-builder-reorg` that have the least downstream impact, yet cause the most merge friction.

There should be few downstream locations to update, and should consist mostly of type signature changes.


what4/src/What4/Expr/App.hs
```diff
@@ -0,0 +1,1861 @@
+{-|
+Module      : What4.Expr.App
+Copyright   : (c) Galois Inc, 2015-2016
+License     : BSD3
+Maintainer  : jhendrix@galois.com
+
+This module defines datastructures that encode the basic
+syntax formers used in What4.ExprBuidler.
```
kquick, line 8: ExprBuilder

what4/src/What4/Expr/App.hs
```diff
@@ -0,0 +1,1861 @@
...
+  FnApp :: !(ExprSymFn t e args ret)
+        -> !(Ctx.Assignment e args)
+        -> NonceApp t e ret
+
+
+------------------------------------------------------------------------
+-- ExprSymFn
+
+-- | This describes information about an undefined or defined function.
+-- Parameter @t@ is a phantom type brand used to track nonces.
```
kquick, line 180: Parameter `e` is new but there's no description of its use.  Adding this would especially help downstream porting efforts.

kquick: I see it described in the NonceApp above... presumably the same thing?

PR #59, opened 2020-07-16
Added UnfoldBoundVars UnfoldStrategy, with isBoundVar in Expr class.

kquick:


what4/src/What4/Interface.hs
```diff
@@ -312,6 +312,9 @@ class HasAbsValue e => IsExpr e where
 
   asAffineVar :: e tp -> Maybe (ConcreteVal tp, e tp, ConcreteVal tp)
 
+  -- | Does the expression reference a Bound Variable (@BoundVar@)?
+  isBoundVar :: e tp -> Bool
```
robdockins, line 316: What about `asBoundVar :: e tp -> Maybe (BoundVar tp)` ?

I think that is a potentially useful operation too.  Then `isBoundVar = isJust . asBoundVar`

kquick: Sure, that sounds reasonable.

kquick: Update: we determined this isn't possible because it's a `BoundVar sym tp`.

robdockins:
@kquick, if you're happy with this, I think we can merge it

kquick:
Per external discussions, I'm closing this without merging.

* The actual semantics around `isBoundVar` are a bit indeterminate regarding a general solution (e.g. perhaps it should be traversing the entire term to find any bound vars in the term).

* A `UserPolicy` can be added (dropping the unused `Eq`, `Ord`, and `Show` instances), but since this is existentially qualified it's not really possible to reify this into a `Builder`-specific case.  Adding the expression type as a parameter (`data UnfoldStrategy e = ...`) is not possible because of the need for a traverse over the `DefinedFnInfo` to change the expr type.

* The original use case that drove the need for this `BoundVar`-related policy has gone away, so there is no demand for this, nor is  there a clear design goal.

PR #61, opened 2020-07-21
solver test updates

kquick:
Misc updates to solver tests for timeouts, reporting, and allowing CI to skip dReal solver tests.


robdockins:
I'm a little confused, I guess.  I intended the `extraTests` flag to remain disabled on CI because it's tricky to set up all the solvers.

robdockins:
Alright, looks good.

For posterity, we decided it was worthwhile to run these solver tests in CI, but to disable the dReal tests for now (because it is a bit tricky to build in our test environment) and the STP tests (because the tests currently stall waiting for models due to an upstream bug).

PR #65, opened 2020-08-14
Add support for Verilog export

atomb:
This PR adds support for translating What4 bitvector terms into the subset of Verilog supported by ABC. The goal is to provide a different mechanism for using ABC as an external process.

This also provides a prover interface for ABC using either Verilog or SMT-Lib2.

There are a couple of lemmas about manipulating `NatRepr`s, in `What4.Protocol.VerilogWriter.Backend` that could perhaps belong in `parameterized-utils`, and could perhaps be better written.


what4/src/What4/Protocol/VerilogWriter/AST.hs
```diff
@@ -0,0 +1,378 @@
...
+  return name
+
+addOutput ::
+  WT.BaseTypeRepr tp ->
+  Identifier ->
+  Exp tp ->
+  VerilogM sym n ()
+addOutput tp name e = do
+  st <- get
+  put $ st { vsOutputs = (Some tp, False, name, Some e) : vsOutputs st }
```
kquick, line 338: `modify`?

what4/src/What4/Protocol/VerilogWriter/Backend.hs
```diff
@@ -0,0 +1,398 @@
+{-
+Module           : What4.Protocol.VerilogWriter.Backend
+Copyright        : (c) Galois, Inc 2020
+Maintainer       : Jennifer Paykin <jpaykin@galois.com>
```
kquick, line 4: Jennifer again.

what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -702,10 +705,11 @@ parseRealSolverValue s = fail $ "Could not parse solver value: " ++ show s
...
 parseBvSolverValue :: MonadFail m => NatRepr w -> SExp -> m (BV.BV w)
 parseBvSolverValue w s
-  | Pair w' bv <- parseBVLitHelper s = case w' `testEquality` w of
-      Just Refl -> return bv
-      Nothing -> fail $ "Solver value parsed with width " ++
-                 show w' ++ ", but should have width " ++ show w
+  | Pair w' bv <- parseBVLitHelper s = case w' `compareNat` w of
+      NatLT zw -> return (BV.zext (addNat w' (addNat zw knownNat)) bv)
+      NatEQ -> return bv
+      NatGT _ -> fail $ "Solver value parsed with width " ++
```
kquick, line 711: Should we use `panic` here instead?

kquick (approved):
Overall looks good, with a few comments about packages and optimizations.

kquick:
We should probably create an issue when this is merged to get `leqSubPos` and `leqSuccLeft` implemented in parameterized-utils and subsequently removed from here, just so these don't get forgotten.

what4/src/What4/Protocol/VerilogWriter.hs
```diff
@@ -0,0 +1,60 @@
+{-# LANGUAGE TypeFamilies #-}
+{-
+Module           : What4.Protocol.VerilogWriter.AST
+Copyright        : (c) Galois, Inc 2020
+Maintainer       : Jennifer Paykin <jpaykin@galois.com>
```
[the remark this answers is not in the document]
atomb: Yeah, she wrote a bunch of this code (when it was in a different repo) and is happy to keep her name on it.

what4/src/What4/Protocol/VerilogWriter.hs
```diff
@@ -0,0 +1,60 @@
...
+-- | Convert the What4 epxression into a Verilog module of the name given
+exprVerilog ::
+  (IsExprBuilder sym, SymExpr sym ~ Expr n) =>
+  sym ->
+  Expr n tp ->
+  Doc () ->
+  ExceptT String IO (Doc ())
+exprVerilog sym e name = fmap (\m -> moduleDoc m name) (exprToModule sym e)
+
+eqVerilog ::
```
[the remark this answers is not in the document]
atomb: Done!

what4/src/What4/Protocol/VerilogWriter.hs
```diff
@@ -0,0 +1,60 @@
...
+import Control.Monad.Except
+import Data.Text.Prettyprint.Doc
+import What4.Expr.Builder (Expr, SymExpr)
+import What4.Interface (IsExprBuilder)
+
+import What4.Protocol.VerilogWriter.AST
+import What4.Protocol.VerilogWriter.ABCVerilog
+import What4.Protocol.VerilogWriter.Backend
+
+-- | Convert the What4 epxression into a Verilog module of the name given
```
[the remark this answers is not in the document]
atomb: Fixed.

what4/src/What4/Protocol/VerilogWriter/ABCVerilog.hs
```diff
@@ -0,0 +1,114 @@
...
+Export Verilog in the particular syntax ABC supports.
+-}
+
+module What4.Protocol.VerilogWriter.ABCVerilog where
+
+import Data.BitVector.Sized
+import Data.Parameterized.NatRepr
+import Data.Parameterized.Some
+import Data.String
+import Data.Text.Prettyprint.Doc
```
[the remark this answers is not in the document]
atomb: Thanks for pointing that out. I've updated it (and one other reference).

what4/src/What4/Protocol/VerilogWriter/ABCVerilog.hs
```diff
@@ -0,0 +1,114 @@
...
+    inputNames = map (identDoc . snd) (vsInputs ms)
+    outputNames = map (identDoc . (\(_, _, n, _) -> n)) (vsOutputs ms)
+    params = reverse inputNames ++ reverse outputNames
+
+typeDoc :: Doc () -> Bool -> BaseTypeRepr tp -> Doc ()
+typeDoc ty _ BaseBoolRepr = ty
+typeDoc ty isSigned (BaseBVRepr w) =
+  ty <+>
+  (if isSigned then "signed " else mempty) <>
+  lbracket <>
```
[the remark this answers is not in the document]
atomb: Ah, good! Fixed.

what4/src/What4/Protocol/VerilogWriter/AST.hs
```diff
@@ -0,0 +1,378 @@
...
+-}
+{-# LANGUAGE GADTs, GeneralizedNewtypeDeriving, ScopedTypeVariables,
+  TypeApplications, PolyKinds, DataKinds, ExplicitNamespaces, TypeOperators #-}
+
+module What4.Protocol.VerilogWriter.AST
+  where
+
+import qualified Data.BitVector.Sized as BV
+import qualified Data.Map as Map
+import           Control.Monad.Except
```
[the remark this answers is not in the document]
atomb: I think I'm going to leave this as-is, based on our conversation.

what4/src/What4/Protocol/VerilogWriter/AST.hs
atomb: Good catch.

what4/src/What4/Protocol/SMTLib2.hs
atomb: Good question. I want to look into how this error gets propagated a little more carefully. It can happy from a poorly-behaved external process, so I'd like to be able to recover gracefully.

atomb:
Yep, that's a good point. I'll make an issue in `parameterized-utils`.

what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -662,7 +665,7 @@ writeCheckSat :: SMTLib2Tweaks a => WriterConn t (Writer a) -> IO ()
 writeCheckSat w = addCommandNoAck w SMT2.checkSat
 
 writeExit :: forall a t. SMTLib2Tweaks a => WriterConn t (Writer a) -> IO ()
-writeExit w = addCommand w SMT2.exit
+writeExit w = mapM_ (addCommand w) (smtlib2exitCommand @a)
```
robdockins, line 668: I don't understand what's happening with this change.

what4/src/What4/Protocol/SMTLib2.hs
robdockins: Yeah, this should probably be an exception instead, here and elsewhere.  But that should maybe be a separate PR.

what4/src/What4/Protocol/VerilogWriter/AST.hs
```diff
@@ -0,0 +1,375 @@
+{-
```
robdockins, line 1: This module could really use a lot more comments.

what4/src/What4/Protocol/VerilogWriter/Backend.hs
```diff
@@ -0,0 +1,398 @@
...
+        where
+          tp = bvarType x
+          base = bvarIdentifier x
+
+eqToVerilogExpr ::
+  (IsExprBuilder sym, SymExpr sym ~ Expr n) =>
+  Expr n tp ->
+  Expr n tp ->
+  VerilogM sym n (IExp WT.BaseBoolType)
+eqToVerilogExpr x y = do
```
robdockins, line 73: This is going to create two separate expression caches and potentially duplicate work.  It is also going to create separate copies of inputs for the same bound variables, which is probably incorrect.

what4/src/What4/Protocol/VerilogWriter/Backend.hs
```diff
@@ -0,0 +1,398 @@
...
+
+appVerilogExpr ::
+  (IsExprBuilder sym, SymExpr sym ~ Expr n) =>
+  App (Expr n) tp ->
+  VerilogM sym n (IExp tp)
+appVerilogExpr app =
+  case app of
+    -- Generic operations
+    BaseIte _ _ b etrue efalse -> do
+      b' <- exprToVerilogExpr b
```
robdockins, line 158: These calls should not be to `exprToVerilogExpr`, as they will each create fresh cache values.  Instead a single cache should be used for the entire evaluation.

robdockins (changes requested):
The Verilog generator doesn't correctly use caching during the process of processing What4 terms.  The result will be a loss of sharing (with potential for exponential term blowup) and possibly spurious counterexamples if inputs are encountered more than once, I think.

Also, we could use quite a bit more comments.

Finally, the git history here is pretty messy.  I'd prefer to see a rebase with some of this stuff cleaned up.

what4/src/What4/Protocol/SMTLib2.hs
atomb: ABC crashes if you give it an `(exit)` command, so we need the option for this list to be empty. It could alternatively be a `Maybe` instead of a list.

what4/src/What4/Protocol/VerilogWriter/AST.hs
atomb: That should be doable!

what4/src/What4/Protocol/VerilogWriter/Backend.hs
atomb: This code actually isn't used anymore (which I guess is why I never saw that blowup happen), so perhaps I'll just delete it.

what4/src/What4/Protocol/SMTLib2.hs
robdockins: Oh, I see. It is already a `Maybe`, but my confusion was not realising that the definition above is a default definition in a class.

what4/src/What4/Protocol/VerilogWriter/Backend.hs
atomb: Oh, good catch!

robdockins: Actually, I just took a closer look... I was mistaken about this, the cache is actually held in the monad state.  Let me take another look to be sure I've understood what's happening.

robdockins: Yeah, whoops, I didn't read things closely enough.  It looks like the cache is being properly piped around via the monad being used, and is only created once in the right place.

atomb:
Looking over the history, @robdockins, I'm not sure there's much value in preserving it. Would you object to squashing when we merge?

atomb:
Oh, one other question: there currently aren't any tests with this, and it's a little tricky to put into CI because it depends on ABC being installed, too. I'm currently putting together some tests for SAW that'll test this code along the way, but would it be worthwhile to include something in the `what4` test suite that at least somewhat covers this code?

robdockins:
I think a squash is OK, as long as PRs elsewhere that depend on this PR via submodules get rebased onto the squashed commit.

It would probably be a good idea to exercise the Verilog generator on at least some simple examples via some golden file testing, yeah.

what4/src/What4/Protocol/VerilogWriter/Backend.hs
atomb: Oh, yeah, the `cache <- vsExpCache <$> get` line isn't creating a new cache.

atomb: I'm still going to get rid of the special function for equality between expressions. I don't think we need it.

Issue #5, opened 2019-04-22
Recover literal floats


langston-barrett:
We have the ability to recover literal values from `SymExpr`s for with `asNat`, `asUnsignedBV`, `asSignedBV`, `asConstantPred`, etc. Is the same thing possible for floats? Could we extract the values of `iFloatLit` back out? This is a complex question due in part to our various possible interpretations for floating point values (uninterpreted, real, IEEE-754).  

Some applications of this would be to pretty-printing and to constant folding.

robdockins:
Fixed via #89

PR #89, opened 2021-01-27
Concrete floats

robdockins:
Use the `libBF` library to implement concrete floating-point ground evaluation and constant folding.

This is especially a tricky area to get right, so I've implemented a new testing technique using "expression templates."   These are fragments of syntax containing variables at base types.  We can enumerate an interesting collection of these templates and use them to generate test cases.  For each test case, we will chose random concrete values for the variables and evaluate the syntax in a variety of ways and test that they agree.  In particular, we can ask solvers to act as computational oracles and verify that our ground evaluation agrees with their theory.


robdockins:
Will fix #5 when merged.

robdockins:
@kquick, This isn't quite ready to merge yet, but I'm interested in your thoughts about the new kind of tests I've been working on (CF https://github.com/GaloisInc/what4/blob/concrete-floats/what4/test/TestTemplate.hs).

Currently they're focused on the floating-point operations, but there's no particular reason the templates idea couldn't be extended to most or all of the main what4 interface.

Issue #91, opened 2021-01-29
structs containing nats don't get represented correctly


benjaminselfridge:
When I create a struct with a single `Nat` field, the constraint that the field is positive does not get output to the solver. I only found this because it resulted in an arithmetic underflow in certain cases.

See https://gist.github.com/benjaminselfridge/b55319660c6dd0609f1c17329abc4cb8

benjaminselfridge:
The same problem also exists for arrays.

I think the problem is here: https://github.com/GaloisInc/what4/blob/master/what4/src/What4/Protocol/SMTWriter.hs#L1225

Perhaps a special case is required for structs and arrays when there is no abstract domain?

robdockins:
Closed via #97

PR #93, opened 2021-01-31
Fixes bug involving structs that contain nats

benjaminselfridge:
This adds the case where we have a struct with no abstract domain, because we
actually need to make a recursive call in that scenario (in case any of the
fields are nats).

This fixes https://github.com/GaloisInc/what4/issues/91

However, it does not address the problem of arrays of nats; I'm not sure how to handle that situation.

### DIFF
--- a/what4/src/What4/Protocol/SMTWriter.hs
+++ b/what4/src/What4/Protocol/SMTWriter.hs
@@ -1221,6 +1221,18 @@ addPartialSideCond ::
 addPartialSideCond _ t NatTypeMap Nothing =
   do addSideCondition "nat_range" $ t .>= 0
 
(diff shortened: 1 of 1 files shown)


robdockins (approved):
This looks correct to me, although it doesn't address the array or function case.

robdockins:
I think I'm going to close this and instead merge #97, which removes `NatType` from the set of base types, and demotes `SymNat` to a derived type.  This prevents nats from appearing in structs, arrays or functions, and seems like the only really reliable way to solve these issues.

benjaminselfridge:
Yep, makes sense to me. Thanks for following up on this PR.

PR #95, opened 2021-02-02
Fix compile with versions >= 4.0

robdockins:
This might be a good time to add realistic version bounds to this part of the code.  Do we have more up-to-date expected versions than are currently in the What4 readme @kquick? Maybe we could put those someplace machine-readable and use TH to automatically import them into this module.

### DIFF
--- a/what4/src/What4/Protocol/SMTLib2.hs
+++ b/what4/src/What4/Protocol/SMTLib2.hs
@@ -74,6 +74,7 @@ module What4.Protocol.SMTLib2
   , checkSolverVersion
   , checkSolverVersion'
   , queryErrorBehavior
+  , defaultSolverBounds
     -- * Re-exports
   , SMTWriter.WriterConn
   , SMTWriter.assume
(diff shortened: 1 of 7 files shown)


what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -1150,8 +1151,8 @@ shutdownSolver _solver p = do
 -----------------------------------------------------------------
 -- Checking solver version bounds
 
-mkChunks :: [Word] -> [Versions.VChunk]
-mkChunks = map ((:[]) . Versions.Digits)
+mkChunks :: Word -> [Word] -> NonEmpty Versions.VChunk
```
kquick, line 1154: I like this, since it's likely the major version number that will be most significant in determining compatibility.

kquick (approved):
The Test portion of the actions now reports the versions that are currently tested (Boolector 3.2.1, CVC4 1.8, Yices 2.6.1, and Z3 4.8.9 where only the latter needs updating in the README).  This a single model within the solution space of supported versions and presumably the  tests could explore that space (and I'd like to with future techniques) and automatically update some table or document but that's probably out of scope for the moment.  I'll create an issue on our DevOps inbox to address this in the future.

As part of this PR you might want to bump Z3 to 4.8.9 in the README.

what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -1151,26 +1152,28 @@ shutdownSolver _solver p = do
...
 --
 -- The keys come from @'smtWriterName'@ in @'WriterConn'@.
 solverMaxVersions :: Map String Version
-solverMaxVersions = []
+solverMaxVersions =
+  [ ( "Yices", $(ver "2.7") )
+  , ( "Z3"   , $(ver "4.9") )
+  , ( "CVC4" , $(ver "1.9") )
+  , ( "STP"  , $(ver "3.3") )
+  ]
```
kquick, line 1176: I'm wondering if we should maintain these min and max table values in an external file that's read in to make it easier to find and/or change.  Then we can also put a link to that file in the README and just maintain the information in a single place.

That would force both of these to be in the IO monad, but currently they are only called from an IO monad location.

robdockins:
https://github.com/GaloisInc/what4/pull/95/checks?check_run_id=1817042175#step:10:114

ugh 👎

what4/src/What4/Protocol/SMTLib2.hs
robdockins: Well, we can do the whole thing in TH, so reading from a file would be no problem, the type of the function wouldn't need to change.

I agree it would be a good idea generally to maintain a canonical place for the info.

kquick: Good point about TH: anyone that will be using this library should have no trouble reading these TH tables. If we put them in their own file for clarity and link to that file from the README I think that keeps all the information clear and locatable.

Is the test failure just the unneeded BV import?

robdockins: No, I think it's a linker error :-(

robdockins:
```
<command line>: libz.so.1: cannot open shared object file: No such file or directory
```

RyanGlScott: I wonder if this is related to the fact that GitHub recently switched over their `ubuntu-latest` runners from using GitHub 18.04 to 20.04. I'm currently debugging a similar issue [here](https://github.com/GaloisInc/saw-script/pull/1057#issuecomment-771916725).

robdockins:
Arrrrggggg

```
Configuring library for config-value-0.8.1..

cabal: The program 'happy' version >=1.19 && <1.21 is required but the version
of
/home/runner/.cabal/store/ghc-8.10.3/happy-1.20.0-a7f2acdd8aee606936f864e4a0252124ecef1a29974252414f914589c5178406/bin/happy
could not be determined.
```

:-( :-( :-(

PR #97, opened 2021-02-03
Demote nat

robdockins:


what4/src/What4/Interface.hs
```diff
@@ -587,6 +682,20 @@ class ( IsExpr (SymExpr sym), HashableF (SymExpr sym)
   -- | Multiply one integer by another.
   intMul :: sym -> SymInteger sym -> SymInteger sym -> IO (SymInteger sym)
 
+  -- | Return the minimum value of two integers
```
benjaminselfridge, line 685: Is the addition of these two functions related to the `SymNat` change?

robdockins: Vaguely.  I've thought we should have them for some time.  `max` is useful for the operation of clamping integers to nonnegative values, which is why I added them here.

PR #100, opened 2021-02-05
Add "getAnnotation" to project SymAnnotations from SymExprs

langston-barrett:
I think the only way to access a `SymAnnotation` embedded in a term is with a traversal. My hope is that the addition of this function makes the annotations API a bit more flexible.


robdockins:
Maybe it makes sense to make this function pure?

langston-barrett:
> Maybe it makes sense to make this function pure?

Certainly easy to do, I was just being cautious/future proof. What's your preference?

robdockins:
I think it's probably better as a pure function.  Then it can be used in pattern guards and such, like the functions that check if values are concrete.

langston-barrett:
> Maybe it makes sense to make this function pure?

Done!

robdockins (approved):

PR #106, opened 2021-03-12
Update cabal version, add some warnings, fix some warnings.

kquick:
Also uses common stanzas in the cabal file which reduces duplication and makes version updates (especially for test modules) easier.


robdockins (approved):
CI is passing, looks OK to me

PR #108, opened 2021-03-12
Remove a couple of SMTReadWriter constraints.

kquick:
These don't seem to be needed, unless they should remain to indicate intent.


robdockins (changes requested):
The `popStackOnly` change seems fine.  For `nameResult` and `versionResult`, we should probably remove the unused first argument too, since there is really no point in having it except to indicate the class instance.

robdockins (approved):

PR #112, opened 2021-04-05
Support Verilog modules with multiple outputs

atomb:
This adds support for generating Verilog modules wiht multiple outputs by allowing the user to pass multiple What4 expressions into the Verilog generation functions.


atomb:
This is actually more limited than what we'll eventually need as it would require all given expressions to have the same type. That'd be fine for translation of vectors, but not for tuples or records.

atomb:
This should do the trick now.

robdockins (approved):

PR #114, opened 2021-04-15
Fix bug in printing Verilog rotations for ABC

atomb:
This code accidentally included Cryptol type annotations in the Verilog output (because the `bv-sized` pretty-printer includes them). Hilariously, we didn't detect this because ABC can still correctly read the result due to its very permissive parser that skips over things it doesn't understand. Attempting to load files containing rotates into Yosys, which includes a stricter Verilog parser, uncovered the error.


robdockins (approved):
I don't really understand this change, but I don't think it can break anything unrelated.

PR #116, opened 2021-04-20
Refactor SMTLib2 response parsing and handling

kquick:
Refactors the response parsing from SMTLib2 solvers:

* Response parsing is more specifically tailored to the valid SMTLib2 solver responses, instead of attempting to use S-Expression parsing for all cases.  
   * Provides better validation of additional response fields that are limited to an enumerated set of strings
   * Allows freedom to parse non-S-Exp responses by parsing directly instead of attempting to extend S-Exp parsing to handle these cases (e.g. `SString`)
   * More extensible for different solver responses
   * Devolves into the original S-Exp parsing result when the more specific parsing fails
   * Removes a significant amount of duplication and consolidates response handling
   * Provides the ability to specify configuration options in the SMTLib2 library (for future use).
   * Adds two new parsing exceptions: `SMTLib2ResponseUnrecognized` and `SMTLib2InvalidResponse`
* Added control over response parsing strictness via a configuration option (`solver.strict_parsing`).  The "strict" parsing mode is roughly equivalent to previous parsing and allows only documented responses.  The "lenient" parsing mode allows intervening lines of unrecognized text before matching the specific response.  This can be very useful for running the solver in verbose mode for developer diagnostics.
* Eliminates duplicate input/output handles stored in multiple places internally
* Adds explicit tests for all the (current) expected solver responses to validate that they are parsed as expected.
* Adds tests for the configuration subsystem
* Adds a `deprecatedOpt` to allow deprecation of an option with possible replacement(s).
* Updates the various solver configs options to prefix with `solver.{SOLVERNAME}.` for a consistent heirarchy.  The older config options are retained and marked as deprecated.
* Modifies the config subsystem to throw explicit exceptions instead of using `fail` for various configuration issues.
* Updates `configHelp` and `getConfigValues` to return the help/value _at_ the named node as well as _below_ the named node (previously only the latter was returned).
* Adds some config helpers for setting via `Some OptionSetting` (as returned by `getOptionSettingFromText`).
* Miscellaneous formatting and wording updates for help and failure messages.

Most of the changes here are expected to remain internal to What4; clients of the What4 library are not expected to need to make many changes to accommodate the changes in this PR.  Potential changes needed:
* Configuration operations throw explicit exceptions instead of `MonadFail`
* User configuration files may need updating and will get deprecation warnings until they do.


what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -1192,45 +1140,40 @@ ppSolverVersionError err =
...
-    tryJust filterAsync (Streams.parseFromStream (parseSExp parseSMTLib2String) s) >>=
-      \case
-        Right (SApp [SAtom ":version", SString v]) -> pure v
-        Right (SApp [SAtom "error", SString msg]) -> throw (SMTLib2Error cmd msg)
-        Right res -> throw (SMTLib2ParseError [cmd] (Text.pack (show res)))
-        Left (SomeException e) ->
-          throwSMTLib2ParseError "version query" cmd e
+versionResult :: WriterConn t a -> IO Text
+versionResult =
+  flip (getLimitedSolverResponse "solver version" $
```
robdockins, line 1171: Nitpick: not a fan of using `flip` like this, I'd rather see a variable binding for the conn.

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -726,6 +735,12 @@ newWriterConn h in_h ack solver_name features bindings cs = do
                        , consumeAcknowledgement = ack
                        }
 
+-- | Strictness level for parsing solver responses
```
robdockins, line 738: Probably would be nice to have a comment on this datatype describing the different modes.

what4/src/What4/Solver/Boolector.hs
```diff
@@ -58,7 +58,7 @@ boolectorOptions =
       executablePathOptSty
       (Just "Path to boolector executable")
       (Just (ConcreteString "boolector"))
-  ]
+  ] <> SMT2.smtlib2Options
```
robdockins, line 61: I don't think we can just add the smtlib2Options to each individual solver like this.  If we install the options for more than one SMTLib2 solver at a time, I think the option will clash and we'll get an error from the `Config` module.

I'm not sure what the best way to deal with this is.  Maybe we should just silently ignore adding options that already exist in the map, if they have the right types.

robdockins (changes requested):
This seems like a really nice refactoring.  My only real concern is about the new strictness option and its behavior if multiple solvers have their options installed at the same time.  We need to figure out what to do about that.

what4/src/What4/Protocol/SMTLib2.hs
kquick: I'm not really either, updated.

what4/src/What4/Solver/Boolector.hs
kquick: OK, this ended up with an overhaul of the configs to allow idempotency and deprecation, and it switched the solver configs to be heirarchical so that it should be easy to have per-solver overrides when we want them.

This made it a big pull request; normally I would keep them smaller and I could split this one up, but it will take a little work to detangle the different sets of changes, so let me know what you would like.

robdockins (approved):
Looks good to me, with a couple of nitpicks.

1) I wonder, do we really need the "solver" prefix for all the solver-specific options? I feel like that might be unnecessary.

2) There are several patches near the end of the PR that are mostly empty and seem to reference the config test module that is only added in the final patch.  Probably worth rebasing those all together.

PR #117, opened 2021-04-29
Avoid spurious failure when grounding NonceApps

danmatichuk:
Use 'mzero' instead of throwing an exception in
'evalGroundNonceApp'. This will be caught and thrown by
'evalGroundExpr', but with a different error message.


robdockins (approved):
Seems like a good change to me.

PR #118, opened 2021-05-07
What4 transition system

Ptival:
This is a tentative merge point for `what4-transition-system`.

I tried to separate the Sally-specific stuff into a `What4.TransitionSystem.Exporter.Sally` module.

However, the transition system is currently heavily influenced by the Sally backend. It's hard to figure out the correct level of abstraction with just one backend, so I think this is likely fine until we need a second backend, at which point the API can be refined to a good middle ground between the backends.


atomb (approved):
I think that your approach here is just right. It clarifies what is explicitly Sally-specific and separates it out. We can refactor the other parts, as necessary, when we have more specific requireements from other back ends.

PR #122, opened 2021-05-14
Support client-provided Verilog input lists

atomb:
This makes it possible for the client of the Verilog writer to specify that a specific set of variables should appear in the module input list in a specific order. This makes it so that the order agrees with what the counterexample processor in SAW expects, and makes Verilog modules intended for use elsewhere able to have inputs that appear in the same order as in a Cryptol or SAWCore function. Previously, inputs were ordered according to their occurrence in the term being exported.

Makes progress in moving away from String along the way.

### DIFF
--- a/what4/src/What4/Protocol/VerilogWriter/AST.hs
+++ b/what4/src/What4/Protocol/VerilogWriter/AST.hs
@@ -284,9 +284,13 @@ data ModuleState sym n =
                 -- ^ All internal wires, in reverse order. Includes the
                 -- type, signedness, name, and initializer of each.
                 , vsSeenNonces :: Map.Map Word64 Identifier
-                -- ^ A map from the identifier nonces seen so far to their identifier names.
+                -- ^ A map from the identifier nonces seen so far to
+                -- their identifier names. These nonces exist for
+                -- identifiers from the original term, but not
+                -- intermediate Verilog variables.
                 , vsUsedIdentifiers :: Set.Set Identifier
-                -- ^ A set of all the identifiers used so far
+                -- ^ A set of all the identifiers used so far, including
+                -- intermediate Verilog variables.
                 , vsExpCache :: IdxCache n IExp
                 -- ^ An expression cache to preserve sharing present in
                 -- the What4 representation.
--- a/what4/src/What4/Protocol/VerilogWriter/AST.hs
+++ b/what4/src/What4/Protocol/VerilogWriter/AST.hs
@@ -273,14 +273,16 @@ data Item where
 data ModuleState sym n =
     ModuleState { vsInputs :: [(Word64, Some WT.BaseTypeRepr, Identifier)]
                 -- ^ All module inputs, in reverse order. We use Word64 for Nonces to avoid GHC bugs.
+                -- For more detail:
+                --   https://gitlab.haskell.org/ghc/ghc/-/issues/2595
+                --   https://gitlab.haskell.org/ghc/ghc/-/issues/10856
+                --   https://gitlab.haskell.org/ghc/ghc/-/issues/16501
                 , vsOutputs :: [(Some WT.BaseTypeRepr, Bool, Identifier, Some Exp)]
                 -- ^ All module outputs, in reverse order. Includes the
                 -- type, signedness, name, and initializer of each.
                 , vsWires :: [(Some WT.BaseTypeRepr, Bool, Identifier, Some Exp)]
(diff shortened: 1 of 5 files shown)


what4/src/What4/Protocol/VerilogWriter/AST.hs
```diff
@@ -265,8 +271,8 @@ data Item where
 -- | Necessary monadic operations
 
 data ModuleState sym n =
-    ModuleState { vsInputs :: [(Some WT.BaseTypeRepr, Identifier)]
-                -- ^ All module inputs, in reverse order.
+    ModuleState { vsInputs :: [(Word64, Some WT.BaseTypeRepr, Identifier)]
+                -- ^ All module inputs, in reverse order. We use Word64 for Nonces to avoid GHC bugs.
```
kquick, line 275: A link or more information on the GHC bugs might be good here, especially if we expect those bugs to be fixed at some point.

what4/src/What4/Protocol/VerilogWriter/AST.hs
```diff
@@ -275,6 +281,10 @@ data ModuleState sym n =
                 -- type, signedness, name, and initializer of each.
                 , vsFreshIdent :: Int
                 -- ^ A counter for generating fresh names.
+                , vsSeenNonces :: Map.Map Word64 Identifier
+                -- ^ A map from the identifier nonces seen so far to their identifier names.
+                , vsUsedIdentifiers :: Set.Set Identifier
+                -- ^ A set of all the identifiers used so far
```
kquick, line 287: Is the set of values in `vsSeenNonces` equivalent to the `vsUsedIdentifiers`?  What's the relationship between these groups of identifiers?

what4/src/What4/Protocol/VerilogWriter/AST.hs
```diff
@@ -366,13 +411,21 @@ addWire tp isSigned name e =
...
+    True ->
+      do let x = vsFreshIdent st
+             freshName = T.concat [basename, "_", fromString (show x)]
+         put $ st { vsFreshIdent = x+1
+                  , vsUsedIdentifiers = Set.insert freshName used
+                  }
+         return freshName
+    False ->
+      do put $ st { vsUsedIdentifiers = Set.insert basename used }
+         return basename
```
kquick, line 428: Could reduce the duplication here:

```
st <- get
let used = vsUsedIdentifiers st
let nm = if basename `Set.member` used
         then T.concat [basename, "_", fromString $ show $ length used]
         else basename
put $ st { vsUsedIdentifiers = Set.insert nm used }
return nm
```
Which would also allow removal of `vsFreshIdent` (assuming this is it's primary use).  The downside is that the differentiation numbers wouldn't be monotonic, but they already skip across different basenames.

kquick (approved):
A couple of suggestions, but overall looks good.

robdockins (approved):

what4/src/What4/Protocol/VerilogWriter/AST.hs
atomb: Good point. I'll add links.

what4/src/What4/Protocol/VerilogWriter/AST.hs
atomb: They are unfortunately not the same, due to `addFreshWire`. The changes in this PR complicate the logic somewhat with the goal that we can use original names as much as possible but still be guaranteed to have unique names everywhere.

what4/src/What4/Protocol/VerilogWriter/AST.hs
atomb: Hmm. This is tempting. I'm a little worried that calculating the length of the used list over and over might be an issue in the case where we have hundreds of thousands of identifiers (which is quite realistic -- one of the main examples I'm testing on right now has around 30k, and it's far from the largest we might consider). That said, using a `Vector` might alleviate that problem (and reduce space usage along the way). I'll experiment with it.

atomb: Oh, wait, it's a `Set` already, which has O(1) size calculation.

what4/src/What4/Protocol/VerilogWriter/AST.hs
kquick: OK.  Maybe consider updating  the comments here to help describe the differences.

PR #123, opened 2021-05-16
Allow versions 5.0

felixonmars:
All tests are passing.


RyanGlScott (approved):
Based on the [`versions-5.0.0` changelog](http://hackage.haskell.org/package/versions-5.0.0/changelog), it doesn't look like any of the changes affect `what4`.

PR #124, opened 2021-05-21
Add a solver adapter for bitwuzla

travitch:
Bitwuzla is a relatively new SMT solver with support for the bitvector, floating
point, array, and uninterpreted function theories.

### DIFF
--- a/what4/src/What4/Solver/Bitwuzla.hs
+++ b/what4/src/What4/Solver/Bitwuzla.hs
@@ -1,0 +1,103 @@
+------------------------------------------------------------------------
+-- |
+-- Module           : What4.Solver.Bitwuzla
+-- Description      : Interface for running Bitwuzla
+-- Copyright        : (c) Galois, Inc 2014-2021
+-- License          : BSD3
+-- Maintainer       : Rob Dockins <rdockins@galois.com>
+-- Stability        : provisional
+--
+-- This module provides an interface for running Bitwuzla and parsing
+-- the results back.
+------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


robdockins:
@travitch, any reason not to just merge this?  Obviously, it's better to have testing set up, but I'm not sure leaving this unmerged is better.

robdockins:
OK, answers that question. I'll leave it alone for now.

what4/src/What4/Solver/Bitwuzla.hs
```diff
@@ -0,0 +1,103 @@
...
+  -> [BoolExpr t]
+  -> (SatResult.SatResult (GroundEvalFn t, Maybe (ExprRangeBindings t)) () -> IO a)
+  -> IO a
+runBitwuzlaInOverride =
+  SMT2.runSolverInOverride Bitwuzla SMT2.nullAcknowledgementAction
+     bitwuzlaFeatures Nothing
+
+instance SMT2.SMTLib2GenericSolver Bitwuzla where
+  defaultSolverPath _ = Process.findSolverPath bitwuzlaPath . WI.getConfiguration
+  defaultSolverArgs _ _ = return ["--smt2", "--model-gen", "--incremental", "--output-format=smt2"]
```
RyanGlScott, line 84: I'm surprised that this doesn't just use the same arguments as in the Boolector configuration, given how similar it is to Bitwuzla. In particular, the absence of `-e=0` seems off, since I'm fairly certain things won't work as intended if `unsat` results cause non-zero exit codes.

kquick (approved):

PR #128, opened 2021-06-10
Enhance timeout handling

kquick:
Updates various solver tests and adds better timeout management.

* Running `cabal test` will only ignore tests for missing solvers unless `CI_TEST_LEVEL` is set and not `0`.  This allows easy development with whatever solver is available in the development environment, but ensures that solvers are present and all tests pass in the CI process.
* Some solvers do not provide the ability to add per-goal timeouts.  Other solvers have some difficulty in honoring that timeout even if it is supported.  This PR adds a "deadman timeout" in What4 itself that will interrupt the solver if it does not voluntarily abort at the designated goal timeout period.
* Modifications to solver error handling to pop frames if possible, but not fail if the solver is not functional enough to allow frame popping.
* Propagate Ctrl-C to solvers to allow solvers to abort processing on Ctrl-C and indicate an incomplete status.
* Adds test for timeouts for online solvers.
* Enables testing the STP solver in CI
* Adds `solver.boolector.timeout` and `solver.stp.timeout` configuration options.
* Internally differentiates between response parsing failures v.s. other failure conditions.
* Various cleanup and response handling fixes and improvements for the corner cases caused by interrupts, exceptions, and failures.

### DIFF
--- a/what4/src/What4/Protocol/Online.hs
+++ b/what4/src/What4/Protocol/Online.hs
@@ -183,11 +183,15 @@ data SolverProcess scope solver = SolverProcess
     -- ^ The amount of time (in seconds) that a solver should spend
     -- trying to satisfy any particular goal before giving up.  A
     -- value of zero indicates no time limit.
-
-    -- THIS IS INFORMATIONAL ONLY AND *NOT* USED TO ACTUALLY SET THE
-    -- SOLVER TIMEOUT!  To set the solver timeout, see the
-    -- solver-specific configuration (e.g. 'z3Timeout', 'cvc4Timeout',
-    -- etc.)
+    --
+    -- Note that it is not sufficient to set just this value to
+    -- control timeouts; this value is used as a reference for common
+    -- code (e.g. SMTLIB2) to determine the timeout for the associated
+    -- timer.  When initialized, this field of the SolverProcess is
+    -- initialized from a solver-specific timeout configuration
+    -- (e.g. z3Timeout); the latter is the definitive reference for
+    -- the timeout, and solver-specific code will likely use the the
+    -- latter rather than this common field.
   }
 
 
--- a/what4/src/What4/Protocol/Online.hs
+++ b/what4/src/What4/Protocol/Online.hs
@@ -47,9 +47,9 @@ import           Control.Monad ( unless )
 import           Control.Monad (void, forM, forM_)
 import           Control.Monad.Catch ( MonadMask, bracket_, onException )
 import           Control.Monad.IO.Class ( MonadIO, liftIO )
+import           Data.IORef
 import           Data.Parameterized.Some
 import           Data.Proxy
-import           Data.IORef
 import           Data.Text (Text)
 import qualified Data.Text.Lazy as LazyText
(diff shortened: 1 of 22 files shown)


what4/src/What4/Protocol/Online.hs
```diff
@@ -169,6 +183,11 @@ data SolverProcess scope solver = SolverProcess
     -- ^ The amount of time (in seconds) that a solver should spend
     -- trying to satisfy any particular goal before giving up.  A
     -- value of zero indicates no time limit.
+
```
robdockins, line 186: Should we just remove this?

what4/src/What4/Protocol/Online.hs
```diff
@@ -440,3 +483,39 @@ getSatResult yp = do
...
+-- Note that this will require re-establishment of the solver process
+-- and any associated context for any subsequent solver goal
+-- evaluation.
+
+withLocalGoalTimeout ::
+  SolverProcess t s
+  -> (WriterConn t s -> IO (SatResult () ()))
+  -> IO (Either SomeException (SatResult () ()))
+withLocalGoalTimeout solverProc action =
+  if getGoalTimeoutInSeconds (solverGoalTimeout solverProc) == 0
```
robdockins, line 502: The comment above says that `solverGoalTimeout` is only advisory?

what4/src/What4/Utils/Process.hs
```diff
@@ -93,6 +93,7 @@ startProcess path args mcwd =
               , std_err = CreatePipe
               , create_group = False
               , cwd = mcwd
+              , delegate_ctlc = True
```
robdockins, line 96: what effect does this end up having?

what4/src/What4/Protocol/Online.hs
kquick: I updated the comment: the original message was from my analysis stage and I forgot to update it based on subsequent activity.  This field is still in use for generic solver code, but it should be set from the solver-specific configuration, and some code will still continue to use the latter.  Let me know if the updated comment is not clear.

what4/src/What4/Protocol/Online.hs
kquick: Fixed the comment as described in the previous note.

what4/src/What4/Utils/Process.hs
kquick: It's not directly related to the timeouts themselves, but it's relative to the more general handling of solvers and solver exceptions that is a part of this PR, and it additionally addresses the user Ctl-C usage.

The effect of this change is that the Ctl-C is now passed on to the solver process to allow that process to abort or do what the solver would normally do on a Ctl-C.  The solver management code should now handle the resulting solver's actions appropriately.   That said, this change could end up with other unintended consequences not revealed during testing, in which case it's easy to revert this single setting and it doesn't invalidate the rest of this PR.

kquick:
Rebased onto current master to resolve merge conflict in tests.

robdockins (approved):
I haven't understood all the details here, but this seems like a good improvement.

PR #129, opened 2021-06-11
Relax the failure conditions for `bvSliceBE` and `bvSliceLE`.

### LINKED ISSUE #113: `bvSliceBE` too restrictive.
CF https://github.com/GaloisInc/saw-script/issues/1166

%< ---------------------------------------------------
  Revision:  8402ae2e14de61eba778f3b56b6de6fb41ece610
  Branch:    HEAD
  Location:  bvSliceBE
  Message:   invalid arguments to slice: 128 0 from vector of length 128
CallStack (from HasCallStack):
  panic, called at src/What4/Panic.hs:17:9 in what4-1.0.0.0.99-inplace:What4.Panic
  panic, called at src/What4/SWord.hs:325:5 in what4-1.0.0.0.99-inplace:What4.SWord
%< ---------------------------------------------------

A slice of 0 bits is rejected by the SWord module with a panic, but it should probably just be accepted instead and return the distinguished `ZBV` value instead.

robdockins:
Previously, these would panic if a 0-bit slice was requested.  Instead,
they now return the distinguished `ZBV` 0-length bitvector value.

Minor other tweaks.

Fixes #113

### DIFF
--- a/what4/src/What4/SWord.hs
+++ b/what4/src/What4/SWord.hs
@@ -314,8 +314,10 @@ bvSliceBE :: forall sym. IsExprBuilder sym => sym
   -> Integer
   -- ^ Starting index, from 0 as most significant bit
   -> Integer
-  -- ^ Number of bits to take (must be > 0)
-  -> SWord sym -> IO (SWord sym)
+  -- ^ Number of bits to take
+  -> SWord sym
+  -> IO (SWord sym)
+bvSliceBE _ _m 0 _  = pure ZBV
 bvSliceBE sym m n (DBV bv)
   | Just (Some nr) <- someNat n,
     Just LeqProof  <- isPosNat nr,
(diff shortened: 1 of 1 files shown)


what4/src/What4/SWord.hs
```diff
@@ -314,8 +314,11 @@ bvSliceBE :: forall sym. IsExprBuilder sym => sym
   -> Integer
   -- ^ Starting index, from 0 as most significant bit
   -> Integer
-  -- ^ Number of bits to take (must be > 0)
-  -> SWord sym -> IO (SWord sym)
+  -- ^ Number of bits to take
+  -> SWord sym
+  -> IO (SWord sym)
+bvSliceBE _ _m 0 _  = pure ZBV
```
kquick, line 320: `_m` could just be `_` for consistency

what4/src/What4/SWord.hs
```diff
@@ -314,8 +314,11 @@ bvSliceBE :: forall sym. IsExprBuilder sym => sym
...
+bvSliceBE _ _ _ ZBV = return ZBV
```
kquick, line 321: Allowing a slice from a ZBV seems invalid, and it doesn't agree with the implementation of `bvAtLE` and `bvAtBE` or with a slice of a non ZBV with invalid index or length... although it does agree with `bvSetBE` and `bvSetLE`

kquick (approved):
Would be great if we had some unit tests for the SWord module, but that's beyond the scope of this PR.

I had some concerns about inconsistency between various operations, but I'll defer to your judgement on whether these are significant or not.

robdockins:
Yeah, I didn't change those `ZBV` cases, but I guess I probably should.  I think it's better to fail fast there.

PR #130, opened 2021-06-11
Faithfully implement the semantics of some string operations

### LINKED ISSUE #103: Out-of-range results for string functions.
The `stringIndexOf` and `stringSubstring` functions are not completely faithful to the SMTLib description of their behaviors for out-of-range inputs.  We should ensure that we properly follow the exact semantics described here https://smtlib.cs.uiowa.edu/theories-UnicodeStrings.shtml

In particular, `stringIndexOf` should return `-1` even for negative and out-of-bounds inputs, and `stringSubstring` returns the "longest substring no longer than n" when "i+n" is longer than the input string, and returns the empty string if `n` is negative or `i` out of range.

robdockins:
Previously, the string index-of and substring functions were
underspecified, whereas the SMTLib theory for these functions
determines precicely their result on out-of-bounds inputs.

Here, we faithfully implement those semantics.

Fixes #103

### DIFF
--- a/what4/src/What4/Expr/Builder.hs
+++ b/what4/src/What4/Expr/Builder.hs
@@ -2782,8 +2782,6 @@ instance IsExprBuilder (ExprBuilder t st fs) where
     | Just x' <- asString x
     , Just off' <- asInteger off
     , Just len' <- asInteger len
-    , 0 <= off', 0 <= len'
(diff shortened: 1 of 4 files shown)


kquick (approved):
Another place it would be great to have some tests... someday.  :-)

Looks good.

PR #131, opened 2021-06-14
V1.2

robdockins:
Updates in preparation for new release.

### DIFF
--- a/what4-transition-system/what4-transition-system.cabal
+++ b/what4-transition-system/what4-transition-system.cabal
@@ -22,7 +22,7 @@ common dependencies
     , mtl
     , parameterized-utils  >=2.0  && <2.2
     , text
-    , what4                ^>=1.1
+    , what4                ^>=1.2
 
 library
   import:           dependencies
--- a/what4/CHANGES.md
+++ b/what4/CHANGES.md
@@ -1,4 +1,29 @@
-# 1.1 (Febuary 2021)
+# 1.2 (June 2021)
+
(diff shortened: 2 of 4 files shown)


robdockins:
Hummm... pretty annoying to have `what4-transition-system` in this repository depend on `language-sally` which depends on `what4`.  Now we have to make coordinated changes in those two repos to bump the version number.

Ptival:
Ah indeed, this sounds annoying. What do you think @atomb ?

atomb:
Ah, this is tricky. Since the Sally MCMT format is essentially an extension of SMT-Lib, `language-sally` uses the What4 notion of SMT-Lib as a core building block.

I guess the easiest way to break the loop would be to leave the `What4.TransitionSystem` module in the `what4` repo, but move the `What4.TransitionSystem.Exporter.Sally` module into the `language-sally` repo. There's a certain logic to that structure.

kquick (approved):
Looks good overall, once you resolve the transition system dependencies.

Ptival:
> I guess the easiest way to break the loop would be to leave the `What4.TransitionSystem` module in the `what4` repo, but move the `What4.TransitionSystem.Exporter.Sally` module into the `language-sally` repo. There's a certain logic to that structure.

I kind of like that, will give it a try.

robdockins:
I think that's a good idea, but let's not hold up this PR/release for it.  For now, I think we can just relax the bounds and bump the submodule.

PR #135, opened 2021-06-16
Include test/responses in hackage tarball

felixonmars:
Fixes test failure when running with the hackage tarball:

```
Test suite solver_parsing_tests: RUNNING...
solver_parsing_tests: test/responses: getDirectoryContents:openDirStream: does not exist (No such file or directory)
Test suite solver_parsing_tests: FAIL
```


robdockins:
Dang.  Thanks for catching this.

PR #141, opened 2021-07-30
Inline the definitions of some utility functions.

robdockins:
This significantly speeds up some workloads.

### DIFF
--- a/what4/src/What4/Protocol/SMTWriter.hs
+++ b/what4/src/What4/Protocol/SMTWriter.hs
@@ -106,7 +106,6 @@ import           Control.Monad.Fail ( MonadFail )
 
 import           Control.Exception
 import           Control.Lens hiding ((.>), Strict)
-import           Control.Monad.Extra
 import           Control.Monad.IO.Class
 import           Control.Monad.Reader
 import           Control.Monad.ST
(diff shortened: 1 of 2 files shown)


what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -787,6 +786,23 @@ cacheLookup
 cacheLookup conn lookup_action =
   readIORef (entryStack conn) >>= firstJustM lookup_action
 
+
+-- | Like 'findM', but also allows you to compute some additional information in the predicate.
+firstJustM :: Monad m => (a -> m (Maybe b)) -> [a] -> m (Maybe b)
+firstJustM _ [] = pure Nothing
+firstJustM p (x:xs) = maybeM (firstJustM p xs) (pure . Just) (p x)
```
kquick, line 793: I normally use `return` in the `Monad` context and only use `pure` in `Applicative` contexts.

kquick (approved):
Minor suggestion above, but looking forward to enjoying this speedup!

robdockins:
Do we expect that the 9.0.1 builds are failing?

RyanGlScott:
I certainly wouldn't have expected them to fail. For reference, both of the test failures look like [this](https://github.com/GaloisInc/what4/pull/141/checks?check_run_id=3204470610#step:12:112):

```
    boolector:                                                                  FAIL
      Exception: user error (Parse exception: Failed reading: empty)
```

Comparing the CI logs from a GHC 8.10.4 job versus a 9.0.1 one, the 8.10.4 job [used Boolector 3.2.1](https://github.com/GaloisInc/what4/pull/141/checks?check_run_id=3204470635#step:12:18), while the 9.0.1 job [used Boolector 3.2.2](https://github.com/GaloisInc/what4/pull/141/checks?check_run_id=3204470610#step:12:18). Moreover, the most recent 9.0.1 CI run on `master` (from 24 days ago) [used Boolector 3.2.1](https://github.com/GaloisInc/what4/runs/3003764104#step:12:18). Perhaps this version difference is to blame?

robdockins:
Seems likely.  I'll try to reproduce locally and open a ticket.

PR #143, opened 2021-07-30
[CI] use nixos-21.05 for GHC 9 tests for boolector 3.2.1

kquick:
The boolector 3.2.2 version causes test failures in the adapter tests
for what4.  This nixos version pin retains boolector 3.2.1 (at least
for now).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,7 @@ jobs:
       - uses: cachix/install-nix-action@v12
         name: Nix unstable base
         # if: matrix.ghc == '9.0.1'
-        if: false
+        if: false  # not currently needed, but may be in the future
         with:
           nix_path: nixpkgs=channel:nixos-unstable
       - uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -35,8 +35,14 @@ jobs:
...
       - uses: cachix/install-nix-action@v12
-        name: Nix unstable base
+        name: Nix 21.05 base
         if: matrix.ghc == '9.0.1'
+        with:
+          nix_path: nixpkgs=channel:nixos-21.05
+      - uses: cachix/install-nix-action@v12
+        name: Nix unstable base
+        # if: matrix.ghc == '9.0.1'
+        if: false
```
RyanGlScott, line 45: Can this `if: false` step be removed?

kquick: I was using that to essentially disable that step, but leave it around since I suspect it will be useful in the future.

kquick: I could also comment out that block if you think that's better.

RyanGlScott: Ah, I wasn't aware that that was intentional. In that case, I'm fine with leaving it in, but it would be worth leaving a comment saying why it's there.

RyanGlScott (approved):

Issue #73, opened 2020-10-22
Z3: structs not being declared, or being redundantly declared when using multi-dimensional arrays


danmatichuk:
When attempting to declare a fresh constant multi-dimensional array, the Z3 backend is not correctly declaring the `Struct2` type needed to index into the array.

```
(declare-fun x!5 () (Array (Struct2 Int (_ BitVec 64)) (_ BitVec 8)))
; (error "line 22 column 29: unknown sort 'Struct2'")
```
Full log: [solver2.log](https://github.com/GaloisInc/what4/files/5425170/solver2.log)

If a redundant term that has the expected struct type is explicitly declared and used, then the backend correctly declares the expected datatype and the array declaration is valid. However, for reasons unknown, Z3 occasionally throws an error saying that the constructor `mk-struct2` has already been declared, but at the point where the type is used.

```
(declare-datatypes (T0 T1) ((Struct2 (mk-struct2 (struct2-proj0 T0) (struct2-proj1 T1)))))
...
(define-fun x!740 () (Array (Struct2 Int (_ BitVec 64)) (_ BitVec 8)) (store x!20 (mk-struct2 1 x!738) x!739))
; (error "line 843 column 54: invalid declaration, function 'mk-struct2' (with the given signature) already declared")
```

Full log : [solver.log](https://github.com/GaloisInc/what4/files/5425168/solver.log)

Z3 version: 4.8.6 - Mac OS X 10.14

robdockins:
Humm... that's pretty strange.  I guess we're messing up the bookeeping regarding which datatypes need to be declared somehow.  Are you doing this in a multithreaded context, perchance?

danmatichuk:
Nope, single-threaded. The second error I find very confusing though and might even be a z3 bug, as it seems to be dependent on which assumption frame you are in. In particular if you don't pop the last frame the `define-fun` works fine.

robdockins:
Do you have a code example that exhibits this behavior?

robdockins:
@danmatichuk, just wanted to ping you again.  Do you have an example, or can you give me more context about when this error happens? I'm not sure how to reproduce.

robdockins:
I think this may have been fixed via #144.  Without a repro case, I'm not sure, but I'm closing this for now.

PR #144, opened 2021-07-30
CVC4 array fix

robdockins:


robdockins:
I think c950c5f may also solve #73, although I don't know how to reproduce the original issue.

PR #147, opened 2021-08-03
String updates

robdockins:
Updates regarding string theories for Z3 and CVC4.


robdockins:
Tests are failing here because the Z3 versions on the build machines (4.8.8 or 4.8.10) is too old.  We need 4.8.11.

robdockins:
@RyanGlScott, do you know enough Nix magic to know how to get newer Z3s for the build environments? I'm out of my depth.

robdockins:
In that case I'll wait for @kquick to have time to look at this.

robdockins:
Included in #177

PR #150, opened 2021-08-09
Add Z3 version matrix and cachix caching.

kquick:


robdockins:
Thanks a lot for working on this Kevin.

I don't fully understand all the implications of this, but it looks like we are running and passing pretty much all tests we intend to except the timeout test for Z3=4.8.12.  Is that right? That feels like a pretty good place to be.

kquick:
> I don't fully understand all the implications of this, but it looks like we are running and passing pretty much all tests we intend to except the timeout test for Z3=4.8.12. Is that right? That feels like a pretty good place to be.

That's right.  There's a big performance regression for 4.8.12 (the time needed by the solver jumps from a couple of seconds to 5-6 minutes, although other than that it seems to work fine) but 4.8.13 and 4.8.14 are back in the faster range, so I don't think we need to worry about that particular box in the functionality matrix.

robdockins:
OK.  Maybe we should just blacklist Z3 4.8.12 then. Either way, looks good to me.

robdockins (approved):

PR #152, opened 2021-08-20
Add arrayCopy, arraySet, and arrayRangeEqual primitives.

andreistefanescu:
The implementation supports single-dimensional arrays with bitvector indices. The encoding in SMTLib uses universal quantification, with Z3 the recommended solver. The operations are undefined if any of the index ranges wraps around.

### DIFF
--- a/what4/src/What4/Expr/Builder.hs
+++ b/what4/src/What4/Expr/Builder.hs
@@ -187,6 +187,7 @@ import qualified Data.BitVector.Sized as BV
 import           Data.Bimap (Bimap)
 import qualified Data.Bimap as Bimap
 import qualified Data.Binary.IEEE754 as IEEE754
+import           Data.Functor.Product
 import           Data.Hashable
 import           Data.IORef
 import           Data.Kind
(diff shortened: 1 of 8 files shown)


what4/src/What4/Expr/Builder.hs
```diff
@@ -1088,11 +1089,57 @@ sbConcreteLookup sym arr0 mcidx idx
     -- Lookups on constant arrays just return value
   | Just (ConstantArray _ _ v) <- asApp arr0 = do
       return v
-    -- Lookups on mux arrays just distribute over mux.
-  | Just (BaseIte _ _ p x y) <- asApp arr0 = do
-      xv <- sbConcreteLookup sym x mcidx idx
-      yv <- sbConcreteLookup sym y mcidx idx
-      baseTypeIte sym p xv yv
+
```
robdockins, line 1092: Note to self: grok this.

what4/src/What4/Expr/Builder.hs
```diff
@@ -1101,8 +1148,95 @@ sbConcreteLookup sym arr0 mcidx idx
...
+        sliceArrayLookupUpdate sym arr lookup_idx
+      _ -> do
+        (sliced_arr, sliced_idx) <- sliceArrayLookupUpdate sym arr lookup_idx
+        sliced_begin_idx <- bvAdd sym begin_idx =<<
+          bvSub sym (Ctx.last sliced_idx) (Ctx.last lookup_idx)
+        sliced_arr <- arraySet sym sliced_arr sliced_begin_idx val len
+        return (sliced_arr, sliced_idx)
+
+    -- Lookups on mux arrays just distribute over mux.
+  | Just (BaseIte _ _ p x y) <- asApp arr0 = do
```
robdockins, line 1217: Another case where exponential slowdown is possible.

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -2616,6 +2616,70 @@ appSMTExpr ae = do
               let expr = ite cond value base_array_value
               SMTName array_type <$> freshBoundFn args resType expr
 
+    CopyArray _w_repr _a_repr dest_arr dest_idx src_arr src_idx len _dest_end_idx _src_end_idx -> do
```
robdockins, line 2619: Note to self: study this.

robdockins (changes requested):
I'm still working on understanding some of the tricker aspects of this, but I'd like to see a bit more documentation in a few places, and I wonder if there is anything preventing us from implementing ground evaluation on these new array constructs.

what4/src/What4/Expr/Builder.hs
andreistefanescu: this is similar to the removed code
```
    -- Lookups on mux arrays just distribute over mux.
  | Just (BaseIte _ _ p x y) <- asApp arr0 = do
      xv <- sbConcreteLookup sym x mcidx idx
      yv <- sbConcreteLookup sym y mcidx idx
      baseTypeIte sym p xv yv
```
with the difference that the mux over arrays results in a lookup with mux over indices and a mux over arrays. I don't think this introduces a new exponential slowdown.

what4/src/What4/Expr/Builder.hs
```diff
@@ -1101,8 +1148,95 @@ sbConcreteLookup sym arr0 mcidx idx
...
+arrayUpdateBase ::
+  ExprBuilder t st fs ->
+  Expr t (BaseArrayType (d::>tp) range) ->
+  IO (Expr t (BaseArrayType (d::>tp) range))
+arrayUpdateBase sym arr0 = case asApp arr0 of
+  Just (UpdateArray _ _ arr _ _) -> arrayUpdateBase sym arr
+  Just (ArrayMap _ _ _ arr) -> arrayUpdateBase sym arr
+  Just (CopyArray _ _ arr _ _ _ _ _ _) -> arrayUpdateBase sym arr
+  Just (SetArray _ _ arr _ _ _ _) -> arrayUpdateBase sym arr
+  Just (BaseIte _ _ p x y) -> do
```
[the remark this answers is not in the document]
andreistefanescu: I suppose turn hash-consing on :-)
I can try to see what happens with the proof without it, but I prefer to open an issue to track this

what4/src/What4/Expr/Builder.hs
robdockins: It's true this behavior was basically already here.  Let's open a ticket to track this, and the other instance I mentioned so we don't forget this is a potential issue.

what4/src/What4/Expr/Builder.hs
```diff
@@ -1088,11 +1089,57 @@ sbConcreteLookup sym arr0 mcidx idx
...
-  | Just (BaseIte _ _ p x y) <- asApp arr0 = do
-      xv <- sbConcreteLookup sym x mcidx idx
-      yv <- sbConcreteLookup sym y mcidx idx
-      baseTypeIte sym p xv yv
+
+    -- A lookup in an array copy is a lookup in the src array when inside the copy range
+  | Just (CopyArray w _a_repr _dest_arr dest_begin_idx src_arr src_begin_idx _len dest_end_idx _src_end_idx) <- asApp arr0
+  , Just (Empty :> (BVIndexLit _ lookup_idx_bv)) <- mcidx
+  , lookup_idx_unsigned <- BV.asUnsigned lookup_idx_bv
+  , Just dest_begin_idx_unsigned <- BV.asUnsigned <$> asBV dest_begin_idx
```
robdockins, line 1097: I don't know if it is worth trying to do this, but you can sometimes get useful answers to range questions using the abstract domain computations even when the inputs you start with aren't fully concrete.

what4/src/What4/Expr/Builder.hs
```diff
@@ -1101,8 +1148,95 @@ sbConcreteLookup sym arr0 mcidx idx
...
+-- from the lookup index.
+sliceArrayLookupUpdate ::
+  ExprBuilder t st fs ->
+  Expr t (BaseArrayType (d::>tp) range) ->
+  Ctx.Assignment (Expr t) (d::>tp) ->
+  IO (Expr t (BaseArrayType (d::>tp) range), Ctx.Assignment (Expr t) (d::>tp))
+sliceArrayLookupUpdate sym arr0 lookup_idx
+  | Just (ArrayMap _ _ entry_map arr) <- asApp arr0 =
+    case asConcreteIndices lookup_idx of
+      Just lookup_concrete_idx ->
```
robdockins, line 1167: Can this case ever occur? I think sbConcreteLookup will already handle the case where this would be useful...

what4/src/What4/Expr/Builder.hs
```diff
@@ -1101,8 +1148,95 @@ sbConcreteLookup sym arr0 mcidx idx
     -- Create select index.
   | otherwise = do
     case exprType arr0 of
-      BaseArrayRepr _ range ->
-        sbMakeExpr sym (SelectArray range arr0 idx)
+      BaseArrayRepr _ range -> do
+        (sliced_arr, sliced_idx) <- sliceArrayLookupUpdate sym arr0 idx
```
robdockins, line 1152: Are you convinced this slicing stuff is worth it?  It breaks sharing pretty badly and is quite complicated, so unless we have evidence that it is really worthwhile in actual usage, I'd be inclined to leave it out.

robdockins:
I'm now relatively convinced about the correctness of these new operations.  The remaining issues I have are:

1) I'd like to see some evidence that the array slicing code is really necessary.  That kind of deep rewriting has caused more problems than it solves, in my past experience.
2) The build failures need addressing.  I think they might actually be related to the haddock comments and not the strictness annotation as the error message suggests.  I don't think haddock for GHC 8.6 and 8.8 can handle comments attached to GADT arguments :-( so I think you should just remove the `^` character and make them non-haddock comments.

Otherwise, I feel pretty good about this.

PR #155, opened 2021-09-03
Add ConcreteFloat constructor

### LINKED ISSUE #137: `ConcreteVal` constructor for floating-point values
The `ConcnreteVal` type has representations for all the base types except floating-point values.  We should correct this oversight.

https://github.com/GaloisInc/what4/blob/41948774604ccc374b99961e2531d48d3d08deae/what4/src/What4/Concrete.hs#L67-L83

RyanGlScott:
I found this to be useful as part of a debugging effort for GaloisInc/crucible#828.

Fixes #137.


kquick (approved):
Looks straightforward.

robdockins (approved):

PR #158, opened 2021-09-22
Fix syntax errors in Verilog variable names

atomb:
Variable names generated by `safeSymbol` can include exclamation marks.
Otherwise, they're guaranteed to be safe variable names in most
languages. Here, we replace `!` with `_` in any names before generating
the final Verilog text.


atomb:
This is, admittedly, a bit of a hack. I'm very open to thoughts on how to fix the issue in a cleaner way, if anyone can think of one.

atomb:
Yeah, I think that's a promising idea, @travitch.

Issue #159, opened 2021-09-22
Special function support for interpreted floating point


robdockins:
In order to represent programs that use trigonometric and complex mathematics, we should add representation support for special functions on real and floating-point values.  Of course, very little solver support for such things exists, but we can at least treat such functions as uninterpreted, or perhaps state useful range restrictions or identities about them.

This will be necessary to, e.g., represent most of the standard C functions from `math.h` and similar constructs in other languages.

robdockins:
At a minimum, we should support standard and hyperbolic trig functions, exponential and log functions.  We might also look at things like gamma, erf, Bessel functions, etc.

robdockins:
PR #160 makes strides in this direction.  I'm not sure when we should call this "done" but this initial work adds support for what i think are the most popular functions.

Currently, the `SpecialFunctions` module exposes information about domain, range and symmetry properties of these functions that are not used anywhere.  Moreover, the IEEE-754 floating point mode and the real mode do not handle any of the special functions. Probably, we should add, at a minimum, support for treating these functions as uninterpreted.

Issue #154, opened 2021-09-02
what4-abc fails to compile due to missing cases for `CopyArray` and friends


RyanGlScott:
After 95fd577a6b7e0d13158da97695c6b7750892f918, `what4-abc` fails to compile:

```
$ cabal build what4-abc
Build profile: -w ghc-8.10.4 -O1
In order, the following will be built (use -v for more details):
 - what4-abc-0.1 (lib) (file src/What4/Solver/ABC.hs changed)
Preprocessing library for what4-abc-0.1..
Building library for what4-abc-0.1..
[1 of 1] Compiling What4.Solver.ABC ( src/What4/Solver/ABC.hs, /home/rscott/Documents/Hacking/Haskell/what4/dist-newstyle/build/x86_64-linux/ghc-8.10.4/what4-abc-0.1/build/What4/Solver/ABC.o, /home/rscott/Documents/Hacking/Haskell/what4/dist-newstyle/build/x86_64-linux/ghc-8.10.4/what4-abc-0.1/build/What4/Solver/ABC.dyn_o )

src/What4/Solver/ABC.hs:292:3: error: [-Wincomplete-patterns, -Werror=incomplete-patterns]
    Pattern match(es) are non-exhaustive
    In a case alternative:
        Patterns not matched:
            CopyArray _ _ _ _ _ _ _ _ _
            SetArray _ _ _ _ _ _ _
            EqualArrayRange _ _ _ _ _ _ _ _ _
    |
292 |   case appExprApp ae of
    |   ^^^^^^^^^^^^^^^^^^^^^...
```

PR #160, opened 2021-09-25
Special functions

robdockins:
Implement a uniform representation for special functions on real and floating-point values.  We still have very limited solver support for these functions, but this allows us to at least represent the majority of the operation appearing in, e.g., `math.h`.

### DIFF
--- a/what4-abc/src/What4/Solver/ABC.hs
+++ b/what4-abc/src/What4/Solver/ABC.hs
@@ -306,18 +306,7 @@ bitblastExpr h ae = do
     SemiRingLe  OrderedSemiRingRealRepr _ _ -> realFail
     RealDiv{} -> realFail
     RealSqrt{} -> realFail
-
-    --------------------------------------------------------------------
-    -- Operations that introduce irrational numbers.
-
-    Pi -> realFail
-    RealSin{} -> realFail
-    RealCos{} -> realFail
-    RealATan2{} -> realFail
-    RealSinh{} -> realFail
-    RealCosh{} -> realFail
-    RealExp{} -> realFail
-    RealLog{} -> realFail
+    RealSpecialFunction{} -> realFail
 
     --------------------------------------------------------------------
     -- Bitvector operations
@@ -487,6 +476,7 @@ bitblastExpr h ae = do
     FloatToSBV{} -> floatFail
     FloatToReal{} -> floatFail
     FloatToBinary{} -> floatFail
+    FloatSpecialFunction{} -> floatFail
 
     ------------------------------------------------------------------------
     -- Array operations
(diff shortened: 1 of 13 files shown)


what4-abc/src/What4/Solver/ABC.hs
```diff
@@ -495,6 +485,9 @@ bitblastExpr h ae = do
     ConstantArray{} -> arrayFail
     SelectArray{} -> arrayFail
     UpdateArray{} -> arrayFail
+    CopyArray{} -> arrayFail
```
RyanGlScott, line 488: I think these cases are enough to fix #154, so it would be worth mentioning that in the commit message.

what4/src/What4/SpecialFunctions.hs
```diff
@@ -0,0 +1,285 @@
...
+
+data FunctionSymmetry r
+  = NoSymmetry
+  | EvenFunction
+  | OddFunction
+ deriving (Show)
+
+data R
+
+-- | Data type for representing
```
RyanGlScott, line 61: Is "Data type for representing" a cut-off sentence?

I was sort of wondering what the criteria was for something to qualify as a special function (e.g., does it suffice to be mentioned in the "Recommended operations" section of the [IEEE 754 standard](https://irem.univ-reunion.fr/IMG/pdf/ieee-754-2008.pdf)?), which might be useful to list here.

what4/src/What4/Expr/App.hs
```diff
@@ -498,6 +494,12 @@ data App (e :: BaseType -> Type) (tp :: BaseType) where
     -> App e (BaseBVType w)
   FloatToReal :: !(e (BaseFloatType fpp)) -> App e BaseRealType
 
+  FloatSpecialFunction ::
```
RyanGlScott, line 497: Out of curiosity, would it make sense to have a similar `SpecialFunction` construct for complex numbers (e.g., to support the operations in `complex.h`)?

RyanGlScott (commented):
This looks sensible to me from a quick glance. Is this enough to fix #159?

what4/src/What4/SpecialFunctions.hs
robdockins: Indeed, I forgot to finish this sentence.

The "recommended operations" list is probably a good place to start, but I imagine we'll add to this collection when and as needed.  On awkward thing I notice about that list is that some of the operation take integer arguments, whereas so far I've only involved real arguments here.

what4/src/What4/Expr/App.hs
robdockins: Good question, I haven't thought too much about how to represent `complex.h`.  It seems to me like the intended implementation there is that `complex` is just a pair of reals in rectangular coordinates, so I'd probably do that instead of directly representing complex floating-point values.

robdockins: Indeed, it seems that clang boils away the handling of complex values into corresponding functions on reals.

PR #165, opened 2021-11-01
Add ceiling and floor special functions

RyanGlScott:
In the same vein as #163.

### DIFF
--- a/what4/src/What4/Expr/Builder.hs
+++ b/what4/src/What4/Expr/Builder.hs
@@ -3348,6 +3348,12 @@ instance IsExprBuilder (ExprBuilder t st fs) where
             | sbFloatReduce sym -> realLit sym (toRational (exp (toDouble c)))
           SFn.Log
             | c > 0, sbFloatReduce sym -> realLit sym (toRational (log (toDouble c)))
+          SFn.Ceiling
+            | c == 0 -> realLit sym 0
+            | sbFloatReduce sym -> realLit sym (toRational (ceiling (toDouble c) :: Integer))
+          SFn.Floor
+            | c == 0 -> realLit sym 0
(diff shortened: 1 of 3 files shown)


robdockins:
I don't think ceiling and floor should be special functions; they should be representable via the `iFloatRound` operation with the appropriate choice of rounding mode, I think.

RyanGlScott:
Ah, I somehow missed `iFloatRound` completely—thanks for the tip. I wonder if we should also have `iFloatCeil` and `iFloatFloor` functions by analogy with `realRound`, `realCeil`, and `realFloor`? Then again, `realCeil` and `reilFloor` don't have default implementations in terms of `realRound`, so perhaps there are some subtleties there that I'm missing.

robdockins:
I guess we should first verify that `iFloatRound` will actually have the exact semantics we want.  If not, we'll have to do something else.

As to the API questions, I guess it makes sense to have a floatFloor and floatCeiling defined in terms of floatRound, if that works out.  For the Real functions, I'm not sure if there's a good reason things are that way, but I think floor and ceiling may have predated realRound.

RyanGlScott:
I've implemented proof-of-concept overrides for the `ceil` and `floor` functions in #905, and based on the unit tests that I've added, they seem to behave as expected. For `ceil`, I chose `RTP` as the rounding mode, and for `floor`, I chose `RTN` as the rounding mode, which is consistent with [this code](https://github.com/GaloisInc/what4/blob/042a8112b1d2b97b91e3399bb976eebe33fc60b8/what4/src/What4/Expr/Builder.hs#L3731-L3732).

I don't really need to have `iFloatCeil` and `iFloatFloor` functions to have this work, although the API discrepancy between `IsExprBuilder` and `IsInterpretedFloatExprBuilder` is a little unusual. In any case, this PR isn't the direction we want to go in, so I'll close this.

PR #167, opened 2021-11-17
Simplify array lookup of update with symbolic index.

andreistefanescu:


what4/src/What4/Expr/Builder.hs
```diff
@@ -1091,6 +1091,19 @@ sbConcreteLookup sym arr0 mcidx idx
   | Just (ConstantArray _ _ v) <- asApp arr0 = do
       return v
 
+  | Just (UpdateArray range idx_tps arr update_idx v) <- asApp arr0
```
travitch, line 1094: Let me suggest that this is complex enough to warrant a high-level comment describing this case

robdockins (approved):
Agree that a comment would be nice.  Otherwise, looks correct to me.

Issue #169, opened 2021-11-24
CI failure on GHC 8.6.5 only (`libz.so: cannot open shared object file: No such file or directory`)


RyanGlScott:
The GHC 8.6.5 job (and _only_ the 8.6.5 job) is failing on `master` currently due to `zlib`-related issues ([example](https://github.com/GaloisInc/what4/runs/4293177190?check_suite_focus=true#step:12:944)):

```
[ 2 of 76] Compiling What4.BaseTypes  ( src/What4/BaseTypes.hs, /home/runner/work/what4/what4/dist-newstyle/build/x86_64-linux/ghc-8.6.5/what4-1.2.1.0.99/build/What4/BaseTypes.o )
<command line>: can't load .so/.DLL for: libz.so (libz.so: cannot open shared object file: No such file or directory)
```

RyanGlScott:
I suspected that this might due to a stale build cache, but even after bumping the cache number in #170, the problem still persists.

RyanGlScott:
Actually, we are already doing just that [here](https://github.com/GaloisInc/what4/blob/dd0755f083e9a611850a5433cfbdd520fd1547ec/.github/workflows/test.yml#L108). Hm.  In that case, I'm stumped. Any idea, @kquick?

RyanGlScott:
Fixed in #171.

PR #170, opened 2021-11-24
Draft: Try bumping cache version (#169)

RyanGlScott:
I'm not exactly sure why GHC wouldn't be able to find `libz.so`, but perhaps building everything from scratch will make a difference. See #169.


RyanGlScott:
The CI still fails on 8.6.5, even after this experiment. This ain't it, chief.

PR #168, opened 2021-11-23
Add Z3 tactic option.

andreistefanescu:

PR #172, opened 2021-12-01
Fix Z3 tactic quoting.

andreistefanescu:


RyanGlScott (approved):
Oops, I should have paid closer attention to that in #168. Thanks!

robdockins (approved):

PR #173, opened 2022-01-05
Allow building with `hashable-1.4.*`

RyanGlScott:
This was prompted by GaloisInc/parameterized-utils#126, but the changes needed on `what4`'s end can be applied independently.

### DIFF
--- a/what4/src/What4/Expr/App.hs
+++ b/what4/src/What4/Expr/App.hs
@@ -937,7 +937,7 @@ instance TestEquality e => TestEquality (NonceApp t e) where
            ]
           )
 
-instance HashableF e => HashableF (NonceApp t e) where
+instance (HashableF e, TestEquality e) => HashableF (NonceApp t e) where
   hashWithSaltF = $(structuralHashWithSalt [t|NonceApp|]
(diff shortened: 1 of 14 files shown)


what4/src/What4/Expr/App.hs
```diff
@@ -937,7 +937,7 @@ instance TestEquality e => TestEquality (NonceApp t e) where
            ]
           )
 
-instance HashableF e => HashableF (NonceApp t e) where
+instance (HashableF e, TestEquality e) => HashableF (NonceApp t e) where
```
robdockins, line 940: Why is this change necessary?

robdockins (approved):
Minor question about one change, but otherwise LGTM.

what4/src/What4/Expr/App.hs
RyanGlScott: The body of this `HashableF` invokes `hashWithSalt` on fields of type `Assignment` (in the `MapOverArrays` and `FnApp` cases). In GaloisInc/parameterized-utils#127, I had to add a `TestEquality` constraint to the `Hashable` instance for `Assignment`, so that constraint must be propagated through to this `HashableF` instance as a result.

PR #177, opened 2022-01-10
Updated strings support with recent CI testing and Z3 compatibility matrix.

kquick:
All the string changes from PR 147, combined with latest master (including running with a full Z3 and GHC matrix) with expected failures for string tests on Z3 prior to 4.8.12.


what4/test/OnlineSolverTest.hs
```diff
@@ -453,7 +452,6 @@ timeoutTests testLevel solvers =
            -- that timeout is reached (i.e. before the race timeout here).
            , let maybeRunTest =
                    case (testSolverName sti, sv) of
-#if !MIN_VERSION_base(4,15,0)
```
robdockins, line 497: Is the internal comment still accurate?

robdockins (approved):

PR #180, opened 2022-01-12
Fix solver versions

kquick:
Runs tests using various fixed versions of the supported solvers.

This now runs all tests against specific revisions of solvers, and will let us expand the set of revisions in the future as needed.  For many of the solvers, this also establishes a "lower limit" to the supported version; if it was important to show support for earlier versions of any of these solvers we could do so, but it would mean additional controls over which tests were run, and so that didn't seem worth it at the present time unless specific needs can be identified.

Overall, this will help us maintain a compatibility matrix and also identify any regressions that occur in the future.

### DIFF
--- a/.github/workflows/gen_matrix.pl
+++ b/.github/workflows/gen_matrix.pl
@@ -60,7 +60,16 @@ solver(cvc4).
 solver(boolector).
 solver(abc).
 
-%% Specify the main version of each variable (i.e. the most interesting value of that variable.
+%% Specify the main version of each variable (i.e. the most
+%% interesting value of that variable.  The main version is the one
+%% used for builds that are testing other variations; there is no
+%% other purpose behind the main_version designation, and there may be
+%% multiple main versions for a particular vector but they currently
+%% provide no advantage.
+%%
+%% The main version selected is usually the "most commonly used"
+%% version (which is not necessarily the "most recently released"
+%% version.
 main_version(ubuntu, "ubuntu-latest").
 main_version(ghc, "8.10.7").
 main_version(z3, "4_8_14").
@@ -70,27 +79,32 @@ main_version(cvc4, "1_8").
 main_version(boolector, "3_2_2").
 main_version(abc, "2021_12_30").
 
-%% Specify the versions of the operating system(s), along with the
-%% main version(s) that are of primary interest.
-version(ubuntu, ubuntu_latest).
+%% Specify the versions of the operating system(s) to use to build and
+%% test the code, the versions of the Haskell compiler to use for
+%% building, and the versions of the solvers to use in tests.
+%%
+%% New versions can be added by simply adding a statement here, and
+%% old versions can be removed by deleting the corresponding
+%% statement.  There is no significance to ordering and the following
+%% can be ordered or grouped in any sequence.
 
-%% Specify the versions of the compiler(s), along with the main
-%% version that is of primary interest.
+version(ubuntu, "ubuntu-latest").
+
 version(ghc, "9.0.2").
 version(ghc, "8.10.7").
 version(ghc, "8.8.4").
 version(ghc, "8.6.5").
 
+version(z3, "4_8_14").
+version(z3, "4_8_13").
+version(z3, "4_8_12").
+version(z3, "4_8_11").
+version(z3, "4_8_10").
 version(z3, "4_8_8").
 version(z3, "4_8_9").
-version(z3, "4_8_10").
-version(z3, "4_8_11").
-version(z3, "4_8_12").
-version(z3, "4_8_13").
-version(z3, "4_8_14").
 
-version(yices, "2_6").
 version(yices, "2_6_1").
+version(yices, "2_6").
 
 version(stp, "2_3_3").
 version(stp, "2_3_2").
@@ -110,12 +124,11 @@ version(abc, "2020_06_22").
 %% No changes normally needed below
 
 matrix(Matrix) :-
-    findall(M, solver_main_compiler_main(M), MainMatrix),
     findall(M,
-            ( solver_all_compiler_main(M)
+            ( solver_main_compiler_main(M)
+            ; solver_all_compiler_main(M)
             ; solver_main_compiler_all(M)
-            ), AuxMatrix),
-    append(MainMatrix, AuxMatrix, FullMatrix),
+            ), FullMatrix),
     Matrix = matrix{ include:FullMatrix }.
 
 main_solvers(Spec) :- findall(X, (solver(S), main_version(S, V) , X=S:V), Spec).
@@ -124,16 +137,16 @@ solver_main_compiler_main(MatrixEntry) :-
     main_solvers(Solvers),
     compiler(Compiler),
     main_version(Compiler, CompilerVersion),
-    os(OS), main_version(OS, OSVersion),
-    MatrixEntry = include{ os:OSVersion }
-                  .put(Compiler, CompilerVersion)
-                  .put(Solvers).
+    matrix_entry(Compiler, CompilerVersion, Solvers, MatrixEntry).
 
 solver_main_compiler_all(MatrixEntry) :-
     main_solvers(Solvers),
     compiler(Compiler),
     version(Compiler, CompilerVersion),
     \+ main_version(Compiler, CompilerVersion),
+    matrix_entry(Compiler, CompilerVersion, Solvers, MatrixEntry).
+
+matrix_entry(Compiler, CompilerVersion, Solvers, MatrixEntry) :-
     os(OS), main_version(OS, OSVersion),
     MatrixEntry = include{ os:OSVersion }
                   .put(Compiler, CompilerVersion)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -18,8 +18,8 @@ views or policies of the Department of Defense or the U.S. Government.
...
-|---------------------------------------|-----|-----------|--------|-------|----------|----------|-----------------|
-| Supported                             | yes | 3.2.2, ?  | 1.8, ? | yes   | 2.3.3, ? | 2.6.4, ? | 4.8.8 -- 4.8.14 |
-| goal timeouts                         | ?   | yes       | yes    | ?     | yes      | yes      | ! 4.8.12        |
-| strings with unicode and escape codes | ?   | ?         | yes    | ?     | ?        | ?        | >= 4.8.12       |
+| Feature                               | ABC | Boolector   | CVC4      | Dreal | STP         | Yices    | Z3                   |
+|---------------------------------------|-----|-------------|-----------|-------|-------------|----------|----------------------|
+| Supported                             | yes | >= 3.2.0, ? | >= 1.8, ? | yes   | >= 2.3.3, ? | 2.6.x, ? | 4.8.8 -- 4.8.14      |
+| goal timeouts                         | ?   | yes         | yes       | ?     | yes         | yes      | ! (4.8.11 or 4.8.12) |
+| strings with unicode and escape codes | ?   | ?           | >= 1.8    | ?     | ?           | ?        | >= 4.8.11            |
```
RyanGlScott, line 25: I wonder if we should make a special mention of Z3 4.8.10 due to GaloisInc/cryptol#1107?

kquick: Is there any way to extract the what4 formula from `negshift.icry` to put into a test here?  I think it would be good to update the matrix, but having a supporting test in this same repo would be even better.

RyanGlScott: https://github.com/Z3Prover/z3/issues/5106#issue-832248620 contains an `.smt2` file that tickles the same issue, but I haven't spent the time to turn this into a What4 expression.

RyanGlScott: On second thought, this wouldn't make for a good `what4` test case anyway. The original Cryptol test case that https://github.com/Z3Prover/z3/issues/5106#issue-832248620 was distilled from made essential use of the SBV backend, and trying the same test case in What4 will not terminate in a reasonable amount of time, regardless of which Z3 version is used. A pity.

kquick: Ahh, OK.

.github/workflows/test.yml
```diff
@@ -37,8 +37,244 @@ jobs:
...
+        stp: [ 2_3_3 ]
+        cvc4: [ 1_8 ]
+        boolector: [ 3_2_2 ]
+        abc: [ 2021_12_30 ]
+        allow-failure: [ false ]
+        # Specify multiple older solver versions for the latest builds
+        # to check backward compatibility.  These older solvers are
+        # not simply added to the general matrix above to avoid
+        # creating excessive numbers of repetitive jobs.
+        include:
```
kquick, line 51: Note for the reviewers: github syntax here is regrettably limited.  Above, we specify the matrix of versions we are interested in.  By default, github will test all combinations in that matrix, but that would create a *very* large target space.  To address this, I've limited the specifications in the above matrix to the current versions of solvers, and then used the `include` section here to include additional target configurations I'd like to test.  I use the `include` to run tests on older solver versions (here I'm testing older solver versions for both GHC 8.10.7 and 9.0.2 since I consider 9.0.2 to still be probationary).

Unfortunately, the syntax for a github `include` requires the explicit specification of all parameters, even if they are going to be the same as in the base matrix above.  Thus, there's a lot of repetition below and some comments describing which field is actually being modified.  It's verbose.  I hate it, but it's the best way I know of.

There's also an `exclude` syntax, which lets you specify which points of the original matrix you do not want tested.  It has the same limitations as the `include`: all parameters must be specified even if they haven't changed.  I believe this route would have yielded even more statements than the minimal matrix + include method I chose.

It's verbose.  I hate it.  I don't

RyanGlScott: I hate this too, and I especially hate the idea of having to maintain this. I'm wondering if we should specify the `include` matrix [as JSON](https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson) instead. That is, have a job that runs before the `linux` job that generates a JSON representation of the `include` matrix as output. Then, the `linux` job can specify the `include` matrix like so:

```yaml
include: ${{fromJson(needs.matrix_prep_job.outputs.matrix)}}
```

See [here](https://stackoverflow.com/a/65434401) for an example of how this can be done. This would allow us to consolidate the logic for specifying the different combinations of solvers into a script of some sort, which seems like a much more maintainable solution in the long run.

kquick: Oh interesting... I wasn't aware of that fromJSON capability. I'll do some noodling on this...

robdockins (commented):
Thanks for working on this Kevin.  I think it's a great step in the right direction.  My only concern is the operational one where we are sometimes running into rate-limiting for the build servers.  I don't know what the right approach to that is.  Maybe only run the full matrix daily rather than on every commit?

robdockins:
@kquick, is there any reason not to merge this while we continue to consider better ways to specify the matrix?

kquick:
Generally no.  I don't think the JSON approach is too difficult though, I just hadn't had any time to work on this.  Since there's a conflict though I'll sign up to resolve the conflict and add the JSON approach by next Monday (if you wanted to merge before then please feel free to).

robdockins:
No urgency here, I just didn't want this to bitrot.

.github/workflows/test.yml
kquick: The matrix is now generated more maintainably.

.github/workflows/gen_matrix.pl
```diff
@@ -0,0 +1,191 @@
...
+version(z3, "4_8_11").
+version(z3, "4_8_12").
+version(z3, "4_8_13").
+version(z3, "4_8_14").
+
+version(yices, "2_6").
+version(yices, "2_6_1").
+
+version(stp, "2_3_3").
+version(stp, "2_3_2").
```
RyanGlScott, line 96: `ghc`, `z3`, and `yices`' version numbers are listed in ascending order, but all other tools (such as `stp` here) use descending order. It would be better to pick one convention and use it consistently.

.github/workflows/gen_matrix.pl
```diff
@@ -0,0 +1,191 @@
...
+solver(stp).
+solver(cvc4).
+solver(boolector).
+solver(abc).
+
+%% Specify the main version of each variable (i.e. the most interesting value of that variable.
+main_version(ubuntu, "ubuntu-latest").
+main_version(ghc, "8.10.7").
+main_version(z3, "4_8_14").
+main_version(yices, "2_6").
```
RyanGlScott, line 67: What criteria are used to determine which version of a tool should be the main one? I had assumed that the most recent versions would be chosen as the main ones, but `yices` and `ghc`'s versions _aren't_ the most recent, unlike for other tools. If there is a reason for this discrepancy, it would be worth noting in the comments.

.github/workflows/gen_matrix.pl
```diff
@@ -0,0 +1,191 @@
...
+main_version(stp, "2_3_3").
+main_version(cvc4, "1_8").
+main_version(boolector, "3_2_2").
+main_version(abc, "2021_12_30").
+
+%% Specify the versions of the operating system(s), along with the
+%% main version(s) that are of primary interest.
+version(ubuntu, ubuntu_latest).
```
RyanGlScott, line 75: This might be my lack of SWI Prolog experience showing, but I'm unclear why the `ubuntu` version used here (`ubuntu_latest`) is different from the corresponding version used above in the `main_version` (`"ubuntu-latest"`, a string literal with a hyphen instead of an underscore).

RyanGlScott (commented):
Fantastic! I'm much happier with the way the CI matrix is handled now. I've left some minor suggestions/questions on the `.pl` script.

.github/workflows/gen_matrix.pl
kquick: Ordering doesn't matter to the code (and I added a comment regarding this) but I do prefer the consistency; thanks for catching this.

.github/workflows/gen_matrix.pl
kquick: Updated the comments to describe this.  Please feel free to re-address this if you'd like more clarity.

.github/workflows/gen_matrix.pl
kquick: Good catch!  Since we always test every configuration against all operating systems (at present) this fact was actually unused, but it would have bitten us in some future update where it became more important.

kquick:
@robdockins I think this is ready for a final review now.

.github/workflows/gen_matrix.pl
RyanGlScott: Thanks. I think the thing in particular I am most confused about is the choice to use an old version of Yices. Can you leave a comment saying why that Yices version is preferred? Moreover, the current CI setup uses Yices 2.6.4, which doesn't appear to be included here (AFAICT).

RyanGlScott (approved):

kquick:
> Thanks for working on this Kevin. I think it's a great step in the right direction. My only concern is the operational one where we are sometimes running into rate-limiting for the build servers. I don't know what the right approach to that is. Maybe only run the full matrix daily rather than on every commit?

This latest version that uses the generated matrix is more efficient (it will vary multiple solvers in a single run since the tests are independent for each). I haven't seen any rate limiting issues in this more recent configuration, but I think the new generator will give us the ability to add customization over time of day or other parameters in the future if needed.

robdockins (approved):

PR #187, opened 2022-02-11
Add `unsafeSetAbstractValue` and `resolveSymBV`

RyanGlScott:
This adds two functions whose functionality are technically separate, but are useful when used in combination with each other. `resolveSymBV` uses an online SMT solver to query the `BVDomain` (i.e., the lower and upper version bounds) of a `SymBV` in situations where `{un,}signedBVBounds` won't suffice. At that point, `unsafeSetAbstractValue` can be used to manually set the `AbstractValue` of a symbolic expression—or, in the case of `SymBV`, manually set the `BVDomain`. I have used this in a downstream project with moderate success, and since the underlying machinery could be applicable in other situations, I thought it worth upstreaming to `what4`.

I have added test cases for both functions to ensure that they behave as expected.

### DIFF
--- a/what4/CHANGES.md
+++ b/what4/CHANGES.md
@@ -30,6 +30,11 @@
     the `Hashable` instance for `SymNat` now has an extra `TestEquality`
     constraint to match its `Eq` instance.
 
+* Add an `unsafeSetAbstractValue` function to the `IsExpr` class which allows
+  one to manually set the `AbstractValue` used in a symbolic expression.
+  As the name suggests, this function is unsound in the general case, so use
+  this with caution.
+
 # 1.2.1 (June 2021)
 
 * Include test suite data in the Hackage tarball.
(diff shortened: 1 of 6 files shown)


RyanGlScott:
Although there are other possible `SearchStrategy` values we could add, I think I'll stop at the two I've added for the time being. I've left a comment describing possible future directions for `SearchStrategy`.

@travitch, how does this look?

travitch:
I added one more question. It looks good to me, I'd just want Rob's opinion with his what4 maintainer hat on

robdockins:
I'll take a look today

robdockins (approved):

RyanGlScott:
The CI passes save for the Z3 4.8.11 jobs, which isn't the fault of this PR. See #189. I'm going to be bold and land this anyway.

PR #190, opened 2022-02-19
remove what4-transition-system dependency on language-sally

Ptival:


Ptival:
Related: https://github.com/GaloisInc/language-sally/pull/8

what4-transition-system/what4-transition-system.cabal
```diff
@@ -17,7 +17,6 @@ common dependencies
     , bytestring
     , containers           ^>=0.6
     , io-streams
-    , language-sally       ^>=0.3
```
RyanGlScott, line 20: Make sure to also remove the references to `language-sally` in the following places:

* https://github.com/GaloisInc/what4/blob/ea717ac94a186b5ee18f138b71d8b4b4b2f00955/cabal.project#L17
* https://github.com/GaloisInc/what4/blob/ea717ac94a186b5ee18f138b71d8b4b4b2f00955/.gitmodules#L10-L12

There's also a test case in `what4-transitions-system` that imports Sally-specific modules:

* https://github.com/GaloisInc/what4/blob/ea717ac94a186b5ee18f138b71d8b4b4b2f00955/what4-transition-system/test/Main.hs#L25
* https://github.com/GaloisInc/what4/blob/ea717ac94a186b5ee18f138b71d8b4b4b2f00955/what4-transition-system/test/Main.hs#L36

I doubt that this still compiles after these changes, so it would be worthwhile to migrate the parts of that test suite that depend on Sally over to `language-sally`.

what4-transition-system/what4-transition-system.cabal
```diff
@@ -28,7 +27,6 @@ library
   import:           dependencies
   exposed-modules:
     What4.TransitionSystem
```
RyanGlScott, line 29: There's another reference to Sally in `What4.TransitionSystem` that is somewhat peculiar:

https://github.com/GaloisInc/what4/blob/ea717ac94a186b5ee18f138b71d8b4b4b2f00955/what4-transition-system/src/What4/TransitionSystem.hs#L83-L87

Is this function still needed now that the Sally-specific bits have been moved out? Or, if is still needed, perhaps the comment should be revised to reflect that it does more than Sally-specific things?

Ptival: Oh yeah, this will still be needed for any output language, but the comment could be less sally-specific.

what4-transition-system/what4-transition-system.cabal
Ptival: Good catch!

PR #197, opened 2022-04-11
Allow building with GHC 9.2

RyanGlScott:
This contains a variety of changes needed to make the libraries in the `what4` compile warning-free with GHC 9.2:

* GHC 9.2 includes `-Woperator-whitespace-ext-conflict` as a part of `-Wall`, which causes `Test.Verification` to warn when built with GHC 9.2. The fix is simple: just add spaces after certain uses of the `$` character.
* GHC 9.2 includes `-Wincomplete-uni-patterns` as a part of `-Wall`, which causes `GenWhat4Expr` to warn when built with GHC 9.2. I fixed the warnings by introducing new `from{Bool,Int,BV8,BV16,BV32,BV64}TestExpr` functions to extract the payloads of different `TestExpr` constructors. This is no less partial than before, but at least this way any accidental partiality will result in a more descriptive error message than `Non-exhaustive patterns in ...`.
* In `base-4.16.*`, `Nat` is now a type synonym for `Natural`, and `GHC.TypeNats` now re-exports `Natural`. This causes an import of `Numeric.Natural` in `What4.SemiRing` (previously required to bring `Natural` into scope) trigger an `-Wunused-imports` warning. I fixed the warning by using explicit imports from `GHC.TypeNats` and `Numeric.Natural`.
* The `aig` and `abcBridge` submodules were bumped to bring in the changes from https://github.com/GaloisInc/aig/pull/12 and https://github.com/GaloisInc/abcBridge/pull/19, respectively, which allows them to build with `base-4.16.*`.
* The upper version bounds on `base` were raised in `what4-abc`, `what4-blt`, and `what4-transition-system` to allow them to build with `base-4.16.*`.

### DIFF
--- a/.github/workflows/gen_matrix.pl
+++ b/.github/workflows/gen_matrix.pl
@@ -90,6 +90,7 @@ main_version(abc, "2021_12_30").
 
 version(ubuntu, "ubuntu-latest").
 
+version(ghc, "9.2.2").
 version(ghc, "9.0.2").
 version(ghc, "8.10.7").
 version(ghc, "8.8.4").
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -98,7 +98,7 @@ jobs:
           case ${{ matrix.ghc }} in
             8.6.5) GHC_NIXPKGS=github:nixos/nixpkgs/20.09 ;;
             9.0.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-unstable ;;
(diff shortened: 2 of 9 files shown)


robdockins (approved):

kquick (approved):
Yeah, the type signatures on hedgehog's `Gen.subtermX` are a bit unfortunate and require the constructor shenanigans.  This change makes the expectations a little more obscured, but as you pointed out it keeps GHC happier and improves the errors in case the developer mixes the wrong statements, so I agree it's making an awkward situation a bit better: thank you for wading through this.

PR #198, opened 2022-04-12
Avoid `testProperty` deprecation warnings with `tasty-hedgehog-1.2`

### LINKED ISSUE #195: New deprecation warning
I don't remember seeing this one before.  I'm not sure what the correct mitigation is yet.

```
test/ExprBuilderSMTLib2.hs:1115:3: warning: [-Wdeprecations]
    In the use of ‘testProperty’ (imported from Test.Tasty.Hedgehog):
    Deprecated: "testProperty will cause Hedgehog to provide incorrect instructions for re-checking properties"
     |
1115 |   testProperty ("test resolveSymBV (" ++ show (PP.pretty searchStrat) ++ ")") $
     |   ^^^^^^^^^^^^

```

RyanGlScott:
This fixes #195—or, at the very least, makes the deprecation warnings that arise from `tasty-hedgehog`'s version of `testProperty` go away. The "real" fix would be to convert every use of `testProperty` into a top-level `Property` value, but given how tedious that would be, I'm disinclined to do this unless someone specifically requests it.

### DIFF
--- a/what4/test/ExprBuilderSMTLib2.hs
+++ b/what4/test/ExprBuilderSMTLib2.hs
@@ -19,7 +19,7 @@ import           ProbeSolvers
 import           Test.Tasty
 import           Test.Tasty.Checklist as TC
 import           Test.Tasty.ExpectedFailure
-import           Test.Tasty.Hedgehog
+import           Test.Tasty.Hedgehog.Alt
 import           Test.Tasty.HUnit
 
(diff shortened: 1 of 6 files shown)


robdockins (approved):
Kind of gross we have to do this, but *shrug*

kquick:
Sorry, lost track of this one, but I agree with Rob: gross and (post-facto) approved.

PR #199, opened 2022-04-14
Add operations for projecting the `SymInteger` from a `SymNat`

### LINKED ISSUE #192: No interface for retrieving concrete values from annotated terms
If a concrete expression is annotated with `annotateTerm`, the resulting expression is no longer considered concrete according to `asConcrete`. This is probably expected (since otherwise those annotations might unexpectedly be simplified away), but the interface doesn't provide any way to retrieve the concrete value again.

One solution is to add a variant of `getAnnotation` that also retrieves the inner, un-annotated term, which would then again be considered concrete according to `asConcrete`.

### LINKED ISSUE #193: Add pure function for getting inner integer from SymNat
`natToInteger` doesn't need to be in the IO monad - it would be convenient if this could be exposed as a pure function instead

robdockins:
and for obtaining the original, unannotated term from an annotation.

Fixes #192
Fixes #193

### DIFF
--- a/what4/src/What4/Interface.hs
+++ b/what4/src/What4/Interface.hs
@@ -120,6 +120,7 @@ module What4.Interface
   , natLe
   , natLt
   , natToInteger
+  , natToIntegerPure
   , bvToNat
   , natToReal
   , integerToNat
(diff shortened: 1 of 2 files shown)


what4/src/What4/Interface.hs
```diff
@@ -675,6 +681,10 @@ class ( IsExpr (SymExpr sym), HashableF (SymExpr sym)
   -- 'annotateTerm' returns the same annotation that 'annotateTerm' did.
   getAnnotation :: sym -> SymExpr sym tp -> Maybe (SymAnnotation sym tp)
 
+  -- | Project the original, unannotated term from an annotation.
```
RyanGlScott, line 684: "from an annotation" sounds a bit strange. I'd suggest something like "from an annotated expression".

RyanGlScott (approved):

Issue #174, opened 2022-01-05
`what4:iteexprs_tests` CI failure


RyanGlScott:
A CI job turned up [this test suite failure](https://github.com/GaloisInc/what4/runs/4718376718?check_suite_focus=true#step:17:339):

```
Test suite iteexprs_tests: RUNNING...
Ite Expressions
    <snip>
    struct mux:                              FAIL (0.90s)
        ✗ struct mux failed 
          after 500 tests.
          and cases         6% █▏·················· ✓ 2%
          or cases         57% ███████████▎········ ✓ 2%
          eq cases          4% ▋··················· ✓ 2%
          xor cases        39% ███████▉············ ✓ 2%
          not cases        43% ████████▌··········· ✓ 2%
          intEq cases       3% ▋··················· ✓ 2%
        ⚠ intLe cases       2% ▎··················· ✗ 2%
          intLt cases       5% █··················· ✓ 2%
          intAdd cases      5% █··················· ✓ 2%
          intSub cases      6% █▏·················· ✓ 2%
          intMul cases      6% █▏·················· ✓ 2%
          intDiv cases      6% █··················· ✓ 2%
          intMod cases      5% █··················· ✓ 2%
          intIte cases      5% █··················· ✓ 2%
          bvCount... cases 62% ████████████▎······· ✓ 2%
        
              ┏━━ test/IteExprs.hs ━━━
          295 ┃ testConcretePredProps :: TestTree
          296 ┃ testConcretePredProps = testGroup "generated concrete predicates" $
          297 ┃   let tt n f = testProperty (n <> " mux") $
          298 ┃                -- withConfidence (10^9) $
          299 ┃ 
          300 ┃                -- increase the # of tests because What4 exprs are
          301 ┃                -- complex and so an increased number of tests is
          302 ┃                -- needed to get reasonable coverage.
          303 ┃                withTests 500 $
          304 ┃ 
          305 ┃                property $ do itc <- forAll genITETestCond
          306 ┃                              -- these cover statements just ensure
          307 ┃                              -- that enough tests have been run to see
          308 ┃                              -- most What4 expression elements.
          309 ┃                              cover 2 "and cases" $ "and" `isInfixOf` (desc itc)
          310 ┃                              cover 2 "or cases" $ "or" `isInfixOf` (desc itc)
          311 ┃                              cover 2 "eq cases" $ "eq" `isInfixOf` (desc itc)
[...]

PR #204, opened 2022-06-28
Adding support for cvc5

### LINKED ISSUE #202: Use `ALL` instead of `ALL_SUPPORTED`?
Currently, `what4`'s [Boolector](https://github.com/GaloisInc/what4/blob/6eda4e395c5fa5110e26ace6b71b11d753788b6f/what4/src/What4/Solver/Boolector.hs#L131) and [CVC4 backends](https://github.com/GaloisInc/what4/blob/719a3b890e661783b504d5380ea39b95d66a9b5c/what4/src/What4/Solver/CVC4.hs#L167) use `ALL_SUPPORTED` as the default logic. I recently did a quick experiment to see if What4's CVC4 settings would work for CVC5, but it turns out that this is not the case, as CVC5 does not recognize `ALL_SUPPORTED`:

```
$ echo "(set-logic ALL_SUPPORTED)" > test.smt2
$ cvc4 test.smt2 
$ cvc5 test.smt2 
(error "Parse Error: test.smt2:1.23: Illegal argument detected

  (set-logic ALL_SUPPORTED)
             ^

void cvc5::internal::LogicInfo::setLogicString(std::string)

  `logicString' is a bad argument
LogicInfo::setLogicString(): junk ("LL_SUPPORTED") at end of logic string: ALL_SUPPORTED")
```

This is a result of cvc5/cvc5#6664, which removes `ALL_SUPPORTED` in favor of `ALL`. Unlike `ALL_SUPPORTED`, `ALL` is actually mentioned by name in the SMT-LIB standard. (See [Section 4.2.1](https://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.6-r2017-07-18.pdf).)

In order to support CVC5, we'll need to add support for `ALL` at the minimum. But I wonder if we shouldn't just use `ALL` everywhere? Being mentioned in the SMT-LIB standard, it's supported by every SMT solver I've tried.

arjunvish:
Adding cvc5 support
Fixes #202

### DIFF
--- a/what4/src/What4/Protocol/SMTLib2/Syntax.hs
+++ b/what4/src/What4/Protocol/SMTLib2/Syntax.hs
@@ -183,7 +183,7 @@ qf_bv = Logic "QF_BV"
 
 -- | Set the logic to all supported logics.
 allSupported :: Logic
-allSupported = Logic "ALL_SUPPORTED"
+allSupported = Logic "ALL" -- TODO RGS: Should ALL_SUPPORTED be a distinct logic?
 
 ------------------------------------------------------------------------
 -- Symbol
@@ -480,7 +480,7 @@ isInt = un_app "is_int"
 -- value type `t2` that always returns `c`.
 --
 -- This uses the non-standard SMTLIB2 syntax
--- @((as const (Array t1 t2)) c)@ which is supported by CVC4 and Z3
+-- @((as const (Array t1 t2)) c)@ which is supported by CVC4, CVC5, and Z3
 -- (and perhaps others).
 arrayConst :: Sort -> Sort -> Term -> Term
 arrayConst itp rtp c =
--- a/what4/src/What4/Protocol/SMTLib2/Syntax.hs
+++ b/what4/src/What4/Protocol/SMTLib2/Syntax.hs
@@ -187,8 +187,9 @@ allSupported :: Logic
 allSupported = Logic "ALL_SUPPORTED"
 {-# DEPRECATED allSupported "Use allLogic instead" #-}
 
+-- | Set the logic to all supported logics.
 allLogic :: Logic
-allLogic = Logic "ALL" -- TODO RGS: Should ALL be a distinct logic?
+allLogic = Logic "ALL"
  
 ------------------------------------------------------------------------
 -- Symbol
--- a/what4/src/What4/Protocol/SMTLib2/Syntax.hs
+++ b/what4/src/What4/Protocol/SMTLib2/Syntax.hs
@@ -53,6 +53,7 @@ module What4.Protocol.SMTLib2.Syntax
   , Logic(..)
   , qf_bv
   , allSupported
+  , allLogic
     -- * Sort
   , Sort(..)
   , boolSort
@@ -183,8 +184,12 @@ qf_bv = Logic "QF_BV"
 
 -- | Set the logic to all supported logics.
 allSupported :: Logic
-allSupported = Logic "ALL" -- TODO RGS: Should ALL_SUPPORTED be a distinct logic?
+allSupported = Logic "ALL_SUPPORTED"
+{-# DEPRECATED allSupported "Use allLogic instead" #-}
 
+allLogic :: Logic
+allLogic = Logic "ALL" -- TODO RGS: Should ALL be a distinct logic?
+ 
 ------------------------------------------------------------------------
 -- Symbol
 
--- a/what4/src/What4/Solver/CVC5.hs
+++ b/what4/src/What4/Solver/CVC5.hs
@@ -151,7 +151,7 @@ writeMultiAsmpCVC5SMT2File sym h ps = do
                 (getOption =<< getOptionSetting RSP.strictSMTParsing cfg)
   c <- SMT2.newWriter CVC5 out_str in_str nullAcknowledgementAction strictness "CVC5"
          True cvc5Features True bindings
-  SMT2.setLogic c SMT2.allSupported
+  SMT2.setLogic c Syntax.allLogic
   SMT2.setProduceModels c True
   forM_ ps $ SMT2.assume c
   SMT2.writeCheckSat c
@@ -183,7 +183,7 @@ instance SMT2.SMTLib2GenericSolver CVC5 where
 
   setDefaultLogicAndOptions writer = do
     -- Tell CVC5 to use all supported logics.
-    SMT2.setLogic writer SMT2.allSupported
+    SMT2.setLogic writer Syntax.allLogic
     -- Tell CVC5 to produce models
     SMT2.setProduceModels writer True
 
@@ -224,7 +224,7 @@ setInteractiveLogicAndOptions writer = do
     when (supportedFeatures writer `hasProblemFeature` useUnsatCores) $ do
       SMT2.setOption writer "produce-unsat-cores" "true"
     -- Tell CVC5 to use all supported logics.
-    SMT2.setLogic writer SMT2.allSupported
+    SMT2.setLogic writer Syntax.allLogic
 
 instance OnlineSolver (SMT2.Writer CVC5) where
   startSolverProcess feat mbIOh sym = do
--- a/what4/src/What4/Solver/CVC5.hs
+++ b/what4/src/What4/Solver/CVC5.hs
@@ -1,8 +1,8 @@
 ------------------------------------------------------------------------
 -- |
 -- Module      : What4.Solver.CVC5
--- Description : Solver adapter code for CVC5
--- Copyright   : (c) Galois, Inc 2015-2020
+-- Description : Solver adapter code for cvc5
+-- Copyright   : (c) Galois, Inc 2022
 -- License     : BSD3
 -- Maintainer  : Rob Dockins <rdockins@galois.com>
 -- Stability   : provisional
@@ -69,7 +69,7 @@ cvc5RandomSeed = configOption knownRepr "solver.cvc5.random-seed"
 cvc5Timeout :: ConfigOption BaseIntegerType
 cvc5Timeout = configOption knownRepr "solver.cvc5.timeout"
 
--- | Control strict parsing for Boolector solver responses (defaults
+-- | Control strict parsing for cvc5 solver responses (defaults
 -- to solver.strict-parsing option setting).
 cvc5StrictParsing :: ConfigOption BaseBoolType
 cvc5StrictParsing = configOption knownRepr "solver.cvc5.strict_parsing"
@@ -128,11 +128,16 @@ instance SMT2.SMTLib2Tweaks CVC5 where
 
 cvc5Features :: ProblemFeatures
 cvc5Features = useComputableReals
+           .|. useLinearArithmetic
+           .|. useNonlinearArithmetic
            .|. useIntegerArithmetic
            .|. useSymbolicArrays
            .|. useStrings
            .|. useStructs
            .|. useFloatingPoint
+           .|. useUnsatCores
+           .|. useUninterpFunctions
+           .|. useDefinedFunctions
            .|. useBitvectors
            .|. useQuantifiers
 
@@ -182,9 +187,9 @@ instance SMT2.SMTLib2GenericSolver CVC5 where
   supportsResetAssertions _ = True
 
   setDefaultLogicAndOptions writer = do
-    -- Tell CVC5 to use all supported logics.
+    -- Tell cvc5 to use all supported logics.
     SMT2.setLogic writer Syntax.allLogic
-    -- Tell CVC5 to produce models
+    -- Tell cvc5 to produce models
     SMT2.setProduceModels writer True
 
 runCVC5InOverride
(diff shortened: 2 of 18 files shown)


what4/src/What4/Protocol/SMTLib2/Syntax.hs
```diff
@@ -184,7 +185,11 @@ qf_bv = Logic "QF_BV"
 -- | Set the logic to all supported logics.
 allSupported :: Logic
 allSupported = Logic "ALL_SUPPORTED"
+{-# DEPRECATED allSupported "Use allLogic instead" #-}
```
RyanGlScott, line 188: It would be worth noting in `what4/CHANGES.md` that `allSupported` is now deprecated in favor `allLogic`.

what4/src/What4/Solver/CVC5.hs
```diff
@@ -0,0 +1,236 @@
...
+--
+-- CVC5-specific tweaks to the basic SMTLib2 solver interface.
+------------------------------------------------------------------------
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TypeApplications #-}
+
+module What4.Solver.CVC5
```
RyanGlScott, line 18: It would be worth mentioning in the changelog that `what4` supports CVC5 now.

what4/src/What4/Solver/CVC5.hs
```diff
@@ -0,0 +1,236 @@
...
+cvc5Path = configOption knownRepr "solver.cvc5.path"
+
+cvc5RandomSeed :: ConfigOption BaseIntegerType
+cvc5RandomSeed = configOption knownRepr "solver.cvc5.random-seed"
+
+-- | Per-check timeout, in milliseconds (zero is none)
+cvc5Timeout :: ConfigOption BaseIntegerType
+cvc5Timeout = configOption knownRepr "solver.cvc5.timeout"
+
+-- | Control strict parsing for Boolector solver responses (defaults
```
RyanGlScott, line 72: There's a typo here:

```suggestion
-- | Control strict parsing for CVC5 solver responses (defaults
```

In fact, the same typo is also present in `What4.Solver.CVC4`. Would you mind updating that file as well?

what4/test/ExprBuilderSMTLib2.hs
```diff
@@ -1191,7 +1191,7 @@ testResolveSymBV searchStrat =
 main :: IO ()
 main = do
   testLevel <- TestLevel . fromMaybe "0" <$> lookupEnv "CI_TEST_LEVEL"
-  let solverNames = SolverName <$> [ "cvc4", "yices", "z3" ]
+  let solverNames = SolverName <$> [ "cvc4", "yices", "z3" ] -- TODO RGS: Add cvc5 here?
```
RyanGlScott, line 1194: Indeed, we might want to follow this TODO's advice and add `cvc5` to this list. I'm not exactly sure if it will work out of the box, but it would be worth a try!

arjunvish:
> 

Not at all, thanks for the comments, they make my work easier :)

RyanGlScott:
The current CI failure is:

```
error: attribute 'cvc5' missing
       at /nix/store/r76czr590rslx3059qrjkgkxbs8741hm-source/flake.nix:180:18:
          179|           };
          180|           cvc5 = pkgs.cvc5 // { # whatever the nixpkgs current version is...
             |                  ^
          181|             v1_0_0  = mkCVC5 "1.0.0";
```

If I had to guess, this is because the CI is using too old of a version of `nixpkgs` to know what `cvc5` is. I think this can be fixed by updating the `nixpkgs` version to a newer one, such as `nixos-22.05`:

https://github.com/GaloisInc/what4/blob/e569514d10f940db20027705388565f3d9f729e3/.github/workflows/test.yml#L42

(@kquick, please correct me if I am wrong here.)

kquick:
That's essentially true, but this will need a two-stage fix.  

First, as @RyanGlScott noted, nixos-22.05 is the first version of nixos that recognizes cvc5.  The nix version that needs to be changed for this is in the GaloisInc/flakes repo.  I believe that switching from the current 21.11 to 22.05 will still work for all older solver versions, so this change should be simple (but this PR will need to use this updated version, which should happen automatically).

Then we have the issue where GHC 8.6.5 was dropped from more recent nixos versions, so we cannot test the combination of GHC 8.6.5 && CVC5.  There are two ways to address this:
 *  The Blue Pill solution is to simply drop GHC 8.6.5 support; that should be reasonable because we have a typical range of 3 GHC versions, and we are ignoring GHC 9.0, but GHC 9.2, GHC 8.10, and GHC 8.8 are probably a reasonable set.  To take this approach simply remove the `version(ghc, '8.6.5').` line from `gen_matrix.pl`.  
 * The Red Pill approach is to setup an exclusion in `gen_matrix.pl` that will cause it to avoid specific combinations.  This is probably something that should be implemented anyhow, but it would involve adding a new prolog fact (e.g. `avoid(ghc, '8.6.5', cvc5, _).`) and then adding a verification that the fact is not present when emitting output.  I'd probably add a `\+ avoid(Compiler, CompilerVersion, Solver, SolverVersion)` clause to `matrix_entry`, but that would involve extracting the latter two from `Solvers`.  This would be useful to have in general if you feel your Prolog skills are up to it, but if that's too much of a distraction, you can just take the Blue Pill.

RyanGlScott:
> Then we have the issue where GHC 8.6.5 was dropped from more recent nixos versions, so we cannot test the combination of GHC 8.6.5 && CVC5.

I'm a bit confused. I thought that the CI was using a _different_ `nixpkgs` version to obtain GHC? Looking at these lines:

https://github.com/GaloisInc/what4/blob/e569514d10f940db20027705388565f3d9f729e3/.github/workflows/test.yml#L97-L104

If I'm reading that correctly, the `nix shell` command is pinning the `nixpkgs` version used for downloading `cabal` + GHC to specific versions, while other tools inherit the default `nixpkgs` version that is provided by `install-nix-action`. Is that not the case?

Assuming that doesn't work, is there any reason why we have to tie the versions of GHC that we use to Nix? Couldn't we just as well download GHC from [`haskell/actions`](https://github.com/haskell/actions) and download the remaining tools from Nix?

kquick:
The snippet you highlighted does cover the GHC, but the solvers are provided via `nix run github:GaloisInc/flakes#cvc5.v${{ matrix.cvc5 }}` which references https://github.com/GaloisInc/flakes/blob/main/flake.nix that contains its own nixpkgs reference.  It's possible to override that with the nixpkgs chosen by this script, but the GaloisInc/flakes should be functional in a standalone manner and that override shouldn't be necessary for an executable.

The nix approach can take better advantage of caching than the haskell/actions approach.  As a rough comparison, the what4 builds take approximately 1m40s to start the actual build step, whereas crucible (excluding additional rust preparation) takes about 4m to reach the same point.  And I'm hoping for even better performance when some pending work lands.

RyanGlScott:
Got it, thanks for the explanation. In that case, I suppose I don't have particularly strong feelings about which of the two options you propose is more desirable. If we went with the approach of excluding CVC5 from GHC 8.6.5 builds, we'd have to add some additional CPP to make sure that the test suite are runnable without CVC5 installed. This would be a bit annoying, but not unprecedented, since we already put similar CPP in place for STP and dReal. See [here](https://github.com/GaloisInc/what4/blob/e569514d10f940db20027705388565f3d9f729e3/what4/what4.cabal#L49-L57).

arjunvish:
It seems to me like neither of you have a preference for one of the options over the other. Since I'm trying to get this integrated as quickly as possible, and since my Prolog skills are pretty poor, I'm leaning on the easy way out, which seems to be option 1. Let me know if either of you have any objections.

what4/test/OnlineSolverTest.hs
```diff
@@ -371,8 +373,9 @@ timeoutTests testLevel solvers =
       -- developer as solvers, What4 formulation, and machine speeds
       -- evolve.
       approxTestTimes :: [ (SolverName, Time) ]
-      approxTestTimes = [ (SolverName "Z3",         2.27 % Second)    -- Z3 4.8.10.  Z3 is good at self timeout.
+      approxTestTimes = [ (SolverName "Z3",         2.7 % Second)    -- Z3 4.8.10.  Z3 is good at self timeout.
```
RyanGlScott, line 376: Why was this changed?

what4/src/What4/Solver/CVC5.hs
```diff
@@ -0,0 +1,241 @@
...
+  smtlib2StructSort tps = Syntax.Sort $ "(Tuple" <> foldMap f tps <> ")"
+    where f x = " " <> Syntax.unSort x
+
+  smtlib2StructCtor args = Syntax.term_app "mkTuple" args
+  smtlib2StructProj _n i x = Syntax.term_app (Syntax.builder_list ["_", "tupSel", fromString (show i)]) [ x ]
+
+cvc5Features :: ProblemFeatures
+cvc5Features = useComputableReals
+           .|. useLinearArithmetic
+           .|. useNonlinearArithmetic
```
RyanGlScott, line 132: These two entries are redundant, as they are [implied by `useIntegerArithmetic` and `useComputableReals`](https://github.com/GaloisInc/what4/blob/e569514d10f940db20027705388565f3d9f729e3/what4/src/What4/ProblemFeatures.hs#L70-L76).

RyanGlScott (commented):
The CI is running now. Hooray! (One job failed due to the same underlying issues as #174, but don't worry about that too much—that's a transient error that should go away on a subsequent run.)

what4/test/OnlineSolverTest.hs
arjunvish: Good catch, I was playing with the numbers and changed it back to the wrong time.

RyanGlScott:
One remaining mystery: for some reason, it now takes much longer than before to run the test suite with these changes. For instance, it took 2m 34s to run the test suite on `master` (see [here](https://github.com/GaloisInc/what4/runs/7115750979?check_suite_focus=true#step:12:1)), but on this PR, it took 1h 18m 58s (see [here](https://github.com/GaloisInc/what4/runs/7141939577?check_suite_focus=true#step:12:1)). I'm not quite sure what is going on, but we should definitely try to figure out what causes this.

RyanGlScott:
I was mistaken: the CI taking an hour to run the tests is _not_ something that was introduced in this PR. I recently restarted the CI jobs on the latest commit in `master` ([here]()), and I observe the same thing as well. I'm forced to conclude that bumping the `nixpkgs` version in the `flakes` repo to 22.05 (see GaloisInc/flakes@57044d2a7ed55d8e4f74bb7dc2ca8e984c6ecce1) is responsible, but I still don't really have a good sense of why that is happening.

RyanGlScott:
I've opened #207 to discuss the matter of the CI slowdown separately.

kquick:
I made a note on #207, but basically I believe this was just due to cache seeding.  Also of note: builds from this PR will be unable to push to cache since the cache key is a repository secret and this PR comes from a fork, rather than a branch on the main repository.

RyanGlScott:
Thanks! I was starting to worry that `nixpkgs` had undergone some horrible performance regression, but you have brought me back to my senses.

what4/test/ExprBuilderSMTLib2.hs
```diff
@@ -1191,7 +1191,7 @@ testResolveSymBV searchStrat =
 main :: IO ()
 main = do
   testLevel <- TestLevel . fromMaybe "0" <$> lookupEnv "CI_TEST_LEVEL"
-  let solverNames = SolverName <$> [ "cvc4", "yices", "z3" ]
+  let solverNames = SolverName <$> [ "cvc4", "cvc5", "yices", "z3" ]
```
RyanGlScott, line 1194: While poking around at the test suites recently, I discovered that adding a new solver to the solver tests alone doesn't do much. In order to actually _use_ CVC5 in this test suite, you'd need to add some tests here:

https://github.com/GaloisInc/what4/blob/cff655cce036c70eb113a2f9b1d98dec8c88c2a4/what4/test/ExprBuilderSMTLib2.hs#L1311-L1313

Perhaps it would suffice to define `cvc5Tests = cvc4Tests` for now. If there are CVC5-specific test cases that we encounter later, we can always add on to `cvc5Tests` in the future.

what4/test/OnlineSolverTest.hs
```diff
@@ -373,6 +375,7 @@ timeoutTests testLevel solvers =
       approxTestTimes :: [ (SolverName, Time) ]
       approxTestTimes = [ (SolverName "Z3",         2.27 % Second)    -- Z3 4.8.10.  Z3 is good at self timeout.
                         , (SolverName "CVC4",       7.5  % Second)    -- CVC4 1.8
+                        , (SolverName "CVC5",       0.40  % Second)   -- CVC5 1.0.0
```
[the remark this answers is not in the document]
arjunvish: Will open an issue for redesigning the challenge problem

arjunvish:
> I made a note on #207, but basically I believe this was just due to cache seeding. Also of note: builds from this PR will be unable to push to cache since the cache key is a repository secret and this PR comes from a fork, rather than a branch on the main repository.

Hm, I was debating whether to push to Ryan's old branch in the repo initially, and admittedly, I didn't think of this consequence. At the very least, I'll keep this in mind for future PRs.

RyanGlScott (approved):
LGTM!

PR #212, opened 2022-07-24
Add support for transition system inputs

chameco:
This PR generalizes the `TransitionSystem` type to include a list of inputs. This is mainly to facilitate a corresponding change to `language-sally`, to support Sally's input variables.

This is a breaking change, but I'm not sure if there are any downstream users to be broken at the moment - if there are, I'd be happy to fix them, so any pointers in that direction would be appreciated!


Ptival:
Seems fairly harmless. I don't know of any downstream user, so happy with the breaking change.

Ptival (approved):

PR #213, opened 2022-08-12
Adding a feature to ask for abducts

arjunvish:
When the SMT solver disproves `g` (finds `~g` to be satisfiable), we can ask for abduct `a` that entail `g`, that is `a ^ ~g` is satisfiable. This feature is available with cvc5.

### DIFF
--- a/.github/workflows/gen_matrix.pl
+++ b/.github/workflows/gen_matrix.pl
@@ -77,7 +77,7 @@ main_version(z3, "4_8_14").
 main_version(yices, "2_6_4").
 main_version(stp, "2_3_3").
 main_version(cvc4, "1_8").
-main_version(cvc5, "1_0_0").
+main_version(cvc5, "1_0_1").
 main_version(boolector, "3_2_2").
 main_version(abc, "2021_12_30").
 
@@ -117,6 +117,7 @@ version(cvc4, "1_8").
 version(cvc4, "1_7").
 
 version(cvc5, "1_0_0").
+version(cvc5, "1_0_1").
 
 version(boolector, "3_2_2").
 version(boolector, "3_2_1").
--- a/cabal.project
+++ b/cabal.project
@@ -6,8 +6,8 @@
 
 packages:
   what4/
-  --what4-abc/
-  --what4-blt/
+  what4-abc/
+  what4-blt/
   what4-transition-system/
 
 optional-packages:
--- a/cabal.project
+++ b/cabal.project
@@ -6,8 +6,8 @@
 
 packages:
   what4/
---  what4-abc/
---  what4-blt/
+  --what4-abc/
+  --what4-blt/
   what4-transition-system/
 
 optional-packages:
--- a/cabal.project
+++ b/cabal.project
@@ -6,8 +6,8 @@
 
 packages:
   what4/
---  what4-abc/
---  what4-blt/
+  what4-abc/
+  what4-blt/
   what4-transition-system/
 
 optional-packages:
--- a/cabal.project
+++ b/cabal.project
@@ -6,8 +6,8 @@
 
 packages:
   what4/
-  what4-abc/
-  what4-blt/
+--  what4-abc/
+--  what4-blt/
   what4-transition-system/
 
 optional-packages:
--- a/what4/README.md
+++ b/what4/README.md
@@ -273,7 +273,7 @@ on as wide a collection of solvers as is reasonable.
 - Z3 versions 4.8.7 through 4.8.12
 - Yices 2.6.1 and 2.6.2
 - CVC4 1.7 and 1.8
-- CVC5 1.0.0 and 1.0.1
+- CVC5 1.0.1
 - Boolector 3.2.1 and 3.2.2
 - STP 2.3.3
     (However, note https://github.com/stp/stp/issues/363, which prevents
--- a/what4/README.md
+++ b/what4/README.md
@@ -273,7 +273,7 @@ on as wide a collection of solvers as is reasonable.
 - Z3 versions 4.8.7 through 4.8.12
 - Yices 2.6.1 and 2.6.2
 - CVC4 1.7 and 1.8
-- CVC5 1.0.0
+- CVC5 1.0.0 and 1.0.1
 - Boolector 3.2.1 and 3.2.2
 - STP 2.3.3
     (However, note https://github.com/stp/stp/issues/363, which prevents
--- a/what4/src/What4/Protocol/Online.hs
+++ b/what4/src/What4/Protocol/Online.hs
@@ -244,7 +244,7 @@ checkSatisfiable proc rsn p =
 
 -- | This is a type representing SMT sorts solely for the purpose of allowing
 --   the user to specify grammars while asking for abducts. 
---   TODO: explore ways of using existing representations of SMT sorts in What4
+--   FIXME: can we use an existing representations of SMT sorts in What4?
 data SMTType = 
     Bool
   | Intgr
--- a/what4/src/What4/Protocol/Online.hs
+++ b/what4/src/What4/Protocol/Online.hs
@@ -38,6 +38,7 @@ module What4.Protocol.Online
   , checkWithAssumptionsAndModel
   , getModel
   , getUnsatCore
+  , getAbduct
   , getUnsatAssumptions
   , getSatResult
   , checkSatisfiable
@@ -235,6 +236,16 @@ checkSatisfiable proc rsn p =
       inNewFrame proc $
         do assume conn p
            check proc rsn
+
+-- | Get an abduct from the SMT solver
+getAbduct :: SMTReadWriter solver => SolverProcess scope solver -> BoolExpr scope -> IO ()
+getAbduct proc t =
+  do let conn = solverConn proc
+     unless (supportedFeatures conn `hasProblemFeature` useProduceAbducts) $
+       fail $ show $ pretty (smtWriterName conn) <+> pretty "is not configured to produce abducts"
+     inNewFrame proc $
+      getSingleAbduct conn t
+      smtAbductResult conn conn
 
 -- | Check if the formula is satisifiable in the current
 --   solver state.  This is done in a
--- a/what4/src/What4/Protocol/Online.hs
+++ b/what4/src/What4/Protocol/Online.hs
@@ -45,8 +45,6 @@ module What4.Protocol.Online
   , getSatResult
   , checkSatisfiable
   , checkSatisfiableWithModel
-  , SMTType(..)
-  , cfgToString
   ) where
 
 import           Control.Concurrent ( threadDelay )
@@ -62,15 +60,10 @@ import           Data.IORef
 #else
 import qualified Data.List as L
 #endif
-import           Data.List (intercalate)
-import qualified Data.Map as Map
-import qualified Data.Set as Set
 import           Data.Parameterized.Some
 import           Data.Proxy
 import           Data.Text (Text)
 import qualified Data.Text.Lazy as LazyText
-import           Data.Parameterized.NatRepr as NatRepr
-import           GHC.TypeNats as TypeNats
 import           Prettyprinter
 import           System.Exit
 import           System.IO
@@ -78,7 +71,6 @@ import qualified System.IO.Error as IOE
 import qualified System.IO.Streams as Streams
 import           System.Process (ProcessHandle, terminateProcess, waitForProcess)
 
-import           What4.BaseTypes
 import           What4.Expr
 import           What4.Interface {-(SolverEvent(..)
                                  , SolverStartSATQuery(..)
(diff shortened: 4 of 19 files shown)


cabal.project
```diff
@@ -6,8 +6,8 @@
 
 packages:
   what4/
-  what4-abc/
-  what4-blt/
+--  what4-abc/
+--  what4-blt/
```
travitch, line 10: Should probably revert this?

what4/src/What4/Protocol/Online.hs
```diff
@@ -236,6 +239,29 @@ checkSatisfiable proc rsn p =
         do assume conn p
            check proc rsn
 
+-- | Get `n` abducts from the SMT solver, the disjunction of which entail `t`, and bind them to `nm`
```
travitch, line 242: What does it mean to "bind" the abducts? My impression is that getting the first abduct binds it to a name and then you have to use the same name in subsequent queries?

Could you make that description a bit more explicit if that is the case?

Does that name persist after getting `n` abducts? Is it scoped to the current frame?

what4/src/What4/Solver/CVC5.hs
```diff
@@ -224,9 +227,14 @@ setInteractiveLogicAndOptions writer = do
...
+    -- FIXME: in cvc5-1.0.1, unsat-cores and produce-abducts are incompatible, but
+    -- cvc5-1.0.2 will fix this, and we can uncomment this once we that is released
     -- Tell cvc5 to compute UNSAT cores, if that feature is enabled
-    when (supportedFeatures writer `hasProblemFeature` useUnsatCores) $ do
-      SMT2.setOption writer "produce-unsat-cores" "true"
+    -- when (supportedFeatures writer `hasProblemFeature` useUnsatCores) $ do
+    --  SMT2.setOption writer "produce-unsat-cores" "true"
+    -- Tell cvc5 to produce abducts, if that feature is enabled
+    when (supportedFeatures writer `hasProblemFeature` useProduceAbducts) $ do
+      SMT2.setOption writer "produce-abducts" "true"
```
travitch, line 237: Can you just add the ticket number for this here?

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -873,6 +873,12 @@ class (SupportTermOps (Term h)) => SMTWriter h where
   -- | Pop 1 existing scope
   popCommand    :: f h -> Command h
 
+  -- | Push 2 new scope
+  push2Command   :: f h -> Command h
+
+  -- | Pop 2 existing scope
+  pop2Command    :: f h -> Command h
```
travitch, line 880: Can you elaborate a bit on the semantics of these? What smtlib is generated for these?

travitch (changes requested):

what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -938,6 +977,20 @@ instance SMTLib2Tweaks a => SMTReadWriter (Writer a) where
...
+  smtAbductResult p s nm t =
+    let abductRsp = \case
+          AckSuccessSExp (SApp (_ : _ : _ : _ : abduct)) -> Just (tail $ init $ sExpToString (SApp abduct))
+          _ -> Nothing
+        cmd = getAbductCommand p nm t
+    in getLimitedSolverResponse "get abduct" abductRsp s cmd
+
+  smtAbductNextResult p s =
+    let abductRsp = \case
+          AckSuccessSExp (SApp (_ : _ : _ : _ : abduct)) -> Just (tail $ init $ sExpToString (SApp abduct))
```
RyanGlScott, line 989: Another occurrence of the tricky `tail $ init` pattern. It would be worth factoring this out into its own function.

what4/test/Abduct.hs
```diff
@@ -0,0 +1,135 @@
...
+
+
+testGetAbducts ::
+  ExprBuilder t st fs ->
+  BoolExpr t ->
+  Int ->
+  IO ()
+testGetAbducts sym f n = do
+  -- Print SMT file in /tmp/
+  mirroredOutput <- openFile "/tmp/what4abduct.smt2" ReadWriteMode
```
RyanGlScott, line 77: It would be worth replacing the hard-coded reference to `/tmp/...` here with [`withSystemTempFile`](https://hackage.haskell.org/package/temporary-1.3/docs/System-IO-Temp.html#v:withSystemTempFile) from the `temporary` library, as the latter is more portable across different operating systems.

what4/test/Abduct.hs
```diff
@@ -0,0 +1,135 @@
...
+-- | Determine whether a predicate is satisfiable, and print out the values of a
+-- set of expressions if a satisfying instance is found.
+prove ::
+  ExprBuilder t st fs ->
+  BoolExpr t ->
+  [(String, IntegerExpr t)] ->
+  IO ()
+prove sym f es = do
+  -- We will use cvc5 to determine if f is satisfiable.
+  mirroredOutput <- openFile "/tmp/what4abductprove.smt2" ReadWriteMode
```
RyanGlScott, line 98: Use [`withSystemTempFile`](https://hackage.haskell.org/package/temporary-1.3/docs/System-IO-Temp.html#v:withSystemTempFile) here.

what4/test/Abduct.hs
```diff
@@ -0,0 +1,135 @@
...
+    putStrLn ""
+  hClose mirroredOutput
+
+testGetAbductOnline ::
+  ExprBuilder t st fs ->
+  [BoolExpr t] ->
+  BoolExpr t ->
+  IO ()
+testGetAbductOnline sym hs g = do
+  mirroredOutput <- openFile "/tmp/what4abductproveonline.smt2" ReadWriteMode
```
RyanGlScott, line 127: Use [`withSystemTempFile`](https://hackage.haskell.org/package/temporary-1.3/docs/System-IO-Temp.html#v:withSystemTempFile) here.

what4/what4.cabal
```diff
@@ -236,6 +236,17 @@ executable quickstart
     parameterized-utils,
     what4
 
+executable abduct
```
RyanGlScott, line 239: Is it also possible to add some test cases for abduction as well? The `abduct` executable here won't be run on CI, so if you want to make sure that this functionality doesn't regress, adding test cases would be the safest bet.

what4/what4.cabal
```diff
@@ -236,6 +236,17 @@ executable quickstart
...
+  main-is: test/Abduct.hs
+  default-language: Haskell2010
+
+  build-depends:
+    base,
+    parameterized-utils,
+    what4,
+    bimap >= 0.2,
```
RyanGlScott, line 247: I don't think this dependency is used anywhere in the `abduct` executable, unless I'm missing something.

what4/src/What4/Protocol/Online.hs
```diff
@@ -236,6 +239,29 @@ checkSatisfiable proc rsn p =
...
+getAbducts ::
+  SMTReadWriter solver =>
+  SolverProcess scope solver ->
+  Int ->
+  String ->
+  BoolExpr scope ->
+  IO [String]
```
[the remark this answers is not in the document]
travitch: We don't have infrastructure to parse this result yet, and it would be a moderate lift to add it. I think it might be better to mark the API as unstable and note that it will change when we are able to add that infrastructure.

what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -891,6 +905,31 @@ runGetValue s e = do
...
+             -> Term
+             -> IO [String]
+runGetAbducts s n nm p = do
+  let nm_t = Text.pack nm
+  let rest = n - 1
+  writeGetAbduct (sessionWriter s) nm_t p
+  replicateM_ rest $ writeGetAbductNext (sessionWriter s)
+  let valRsp = \case
+        -- SMT solver returns `(define-fun nm () Bool X)` where X is the abduct, we discard everything but the abduct
+        AckSuccessSExp (SApp (_ : _ : _ : _ : abduct)) -> Just (tail $ init $ sExpToString (SApp abduct))
```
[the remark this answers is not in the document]
travitch: Is it maybe removing quotes or an outer layer of parentheses?

what4/src/What4/Solver/Yices.hs
```diff
@@ -492,6 +495,8 @@ instance SMTWriter Connection where
 
   pushCommand _   = const $ safeCmd "(push)"
   popCommand _    = const $ safeCmd "(pop)"
+  push2Command _   = const $ safeCmd "(push)"
+  pop2Command _    = const $ safeCmd "(pop)"
```
travitch, line 499: This is different than the implementation for smtlib2 (i.e., I don't see the parameter) - is this something that yices just doesn't support? It would probably be better to make it an error (even if we just aren't sure it is supported) than to implement it in a way that might be very surprising

arjunvish: Yeah Yices the Yices syntax doesn't name assertion frames using numerals, although I think it mostly works the same way without numerals. You're right though, best to make it an error.

what4/src/What4/Protocol/SMTWriter.hs
arjunvish: Seems as descriptive as the [previous](https://github.com/GaloisInc/what4/blob/99a6808c0ebc0907e1ec584a5089e72de18dd4fc/what4/src/What4/Protocol/SMTWriter.hs#L870) two commands. I do explain [here](https://github.com/GaloisInc/what4/blob/99a6808c0ebc0907e1ec584a5089e72de18dd4fc/what4/src/What4/Protocol/Online.hs#L356), but I can move/copy this if you think it might be more appropriate in SMTWriter.

what4/README.md
```diff
@@ -273,7 +273,7 @@ on as wide a collection of solvers as is reasonable.
 - Z3 versions 4.8.7 through 4.8.12
 - Yices 2.6.1 and 2.6.2
 - CVC4 1.7 and 1.8
-- CVC5 1.0.0
+- CVC5 1.0.0 and 1.0.1
```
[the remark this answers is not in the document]
arjunvish: What does the `?` mean there? For instance, what does it mean for `>= 1.8, ?` to be supported for cvc4?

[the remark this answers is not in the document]
RyanGlScott: Hah, I think that's simply an indication that we aren't entirely sure what the lower bounds are. For CVC5, we have a much better idea of what the bounds are, so we can simply forgo the use of question marks in the CVC5 column.

what4/src/What4/Protocol/SMTLib2.hs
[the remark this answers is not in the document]
arjunvish: It was a lazy solution to remove parentheses :). I agree, I'll change it.

what4/what4.cabal
RyanGlScott: Alternatively, perhaps we should just convert this into a test suite. It doesn't contain regression tests per se, but it does at least ensure that all of the relevant machinery works without falling over, which is no small feat.

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -867,12 +867,18 @@ class (SupportTermOps (Term h)) => SMTWriter h where
...
 
-  -- | Pop 1 existing scope
+  -- | Generates command `(pop 1)` that closes the corresponding assertion frame
   popCommand    :: f h -> Command h
 
+  -- | Generates command `(push 2)` that opens the corresponding assertion frame
+  push2Command   :: f h -> Command h
+
+  -- | Generates command `(pop 2)` that closes the corresponding assertion frame
+  pop2Command    :: f h -> Command h
```
RyanGlScott, line 880: State that these are currently only used for abduction purposes.

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -867,12 +867,18 @@ class (SupportTermOps (Term h)) => SMTWriter h where
...
 
-  -- | Push 1 new scope
+  -- | Generates command `(push 1)` that opens the corresponding assertion frame
   pushCommand   :: f h -> Command h
 
-  -- | Pop 1 existing scope
+  -- | Generates command `(pop 1)` that closes the corresponding assertion frame
   popCommand    :: f h -> Command h
 
+  -- | Generates command `(push 2)` that opens the corresponding assertion frame
```
RyanGlScott, line 876: A minor thing: Haddock uses the `@` character to delimit inline code, not a backtick:

```suggestion
  -- | Generates command @(push 2)@ that opens the corresponding assertion frame
```

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -867,12 +867,18 @@ class (SupportTermOps (Term h)) => SMTWriter h where
...
+  push2Command   :: f h -> Command h
+
+  -- | Generates command `(pop 2)` that closes the corresponding assertion frame
```
RyanGlScott, line 879:
```suggestion
  -- | Generates command @(pop 2)@ that closes the corresponding assertion frame
```

.github/workflows/gen_matrix.pl
```diff
@@ -117,6 +117,7 @@
 version(cvc4, "1_7").
 
 version(cvc5, "1_0_0").
```
RyanGlScott, line 119: Remove this version, since we don't want to test it.

what4/src/What4/Protocol/Online.hs
```diff
@@ -236,6 +239,39 @@ checkSatisfiable proc rsn p =
         do assume conn p
            check proc rsn
 
+-- | `get-abuct nm t` queries the solver for the first abduct, which is returned
+--   as an SMT function definition named `nm`. The remaining abducts are obtained
+--   from the solver by successive invocations of the `get-abduct-next` command,
+--   which return SMT functions bound to the same `nm` as the first. The name `nm`
```
RyanGlScott, line 245:
```suggestion
-- | @get-abuct nm t@ queries the solver for the first abduct, which is returned
--   as an SMT function definition named @nm@. The remaining abducts are obtained
--   from the solver by successive invocations of the @get-abduct-next@ command,
--   which return SMT functions bound to the same @nm@ as the first. The name @nm@
```

what4/src/What4/Protocol/Online.hs
```diff
@@ -327,6 +363,17 @@ tryPop p =
 inNewFrame :: (MonadIO m, MonadMask m, SMTReadWriter solver) => SolverProcess scope solver -> m a -> m a
 inNewFrame p action = inNewFrameWithVars p [] action
 
+-- For abduction, we want the final assertion to be a in a new frame, so that it 
+-- can be closed before asking for abducts. The following two commands allow frame 2 
+-- to be pushed and popped independently of other commands
+-- | Open a second solver assumption frame.
```
RyanGlScott, line 369: It would be worth moving the `-- | Open a second solver assumption frame.` part to the top of the comments so that the commentary about abduction is included in the Haddocks.

RyanGlScott (commented):
Sorry for taking so long to get back to this.

what4/src/What4/Solver/CVC5.hs
```diff
@@ -224,9 +227,15 @@ setInteractiveLogicAndOptions writer = do
     SMT2.setOption writer "produce-models" "true"
     -- Tell cvc5 to make declarations global, so they are not removed by 'pop' commands
     SMT2.setOption writer "global-declarations" "true"
+    -- FIXME: in cvc5-1.0.1, unsat-cores and produce-abducts are incompatible, but
+    -- cvc5-1.0.2 will fix this, and we can uncomment this once we that is released.
+    -- Closing issue #214 will fix this.
```
RyanGlScott, line 232: CVC5 1.0.2 was [released](https://github.com/cvc5/cvc5/releases/tag/cvc5-1.0.2) recently. We should check to see if that version fixes the issues observed here.

arjunvish:
>

what4/what4.cabal
arjunvish: Agreed, working on converting it into a series of tests.

what4/test/Abduct.hs
```diff
@@ -0,0 +1,158 @@
...
+  IO [String]
+testGetAbductOnline sym hs g n = do
+  -- Print SMT file in /tmp/
+  withSystemTempFile "what4abdonline" $ \fname mirroredOutput -> do
+    proc <- startSolverProcess @(SMT2.Writer CVC5) cvc5Features (Just mirroredOutput) sym
+    let conn = solverConn proc
+    inNewFrame proc $ do
+      mapM_ (\x -> assume conn x) hs
+      getAbducts proc n (pack "abd") g
+    -- hClose mirroredOutput
```
RyanGlScott, line 53: Can this line be removed? I would expect this to be unnecessary since `withSystemTempFile` should close the handle for you.

what4/test/Abduct.hs
```diff
@@ -0,0 +1,158 @@
...
+  -- Print SMT file in /tmp/
+  withSystemTempFile "what4abdoffline" $ \fname mirroredOutput -> do
+    let logData = LogData { logCallbackVerbose = \_ _ -> return ()
+                           , logVerbosity = 2
+                           , logReason = "defaultReason"
+                           , logHandle = Just mirroredOutput }
+    withCVC5 sym cvc5executable logData $ \session -> do
+      f_term <- mkSMTTerm (sessionWriter session) f
+      runGetAbducts session n (pack "abd") f_term
+    --hClose mirroredOutput
```
RyanGlScott, line 71: Similarly, can this line be removed?

what4/test/Abduct.hs
```diff
@@ -0,0 +1,158 @@
...
+  BoolExpr t ->
+  Int ->
+  IO [String]
+testGetAbductOffline sym f n = do
+  -- Print SMT file in /tmp/
+  withSystemTempFile "what4abdoffline" $ \fname mirroredOutput -> do
+    let logData = LogData { logCallbackVerbose = \_ _ -> return ()
+                           , logVerbosity = 2
+                           , logReason = "defaultReason"
+                           , logHandle = Just mirroredOutput }
```
RyanGlScott, line 67: Minor formatting suggestion:

```suggestion
    let logData = LogData { logCallbackVerbose = \_ _ -> return ()
                          , logVerbosity = 2
                          , logReason = "defaultReason"
                          , logHandle = Just mirroredOutput }
```

what4/test/Abduct.hs
```diff
@@ -0,0 +1,158 @@
...
+-- Tests
+
+testAbdOnline :: ExprBuilder t st fs -> 
+  [BoolExpr t] -> 
+  BoolExpr t -> 
+  TestTree
+testAbdOnline sym hs g = testCase "getting 3 abducts using cvc5 online" $ do
+  -- Ask for 3 abducts for f
+  res <- testGetAbductOnline sym hs g 3
+  (Data.List.length res == 3) @? "3 online abducts"
```
RyanGlScott, line 102: Any reason not to just write `length` here instead of `Data.List.length`?

what4/test/Abduct.hs
```diff
@@ -0,0 +1,158 @@
...
+
+testAbdOffline :: ExprBuilder t st fs -> 
+  BoolExpr t -> 
+  [(String, IntegerExpr t)] -> 
+  TestTree
+testAbdOffline sym f es = testCase "getting 3 abducts using cvc5 offline" $ do
+  -- Ask for 3 abducts for f
+  res <- testGetAbductOffline sym f 3
+  (Data.List.length res == 3) @? "3 offline abducts"
```
RyanGlScott, line 111: Any reason not to just write `length` here instead of `Data.List.length`?

what4/test/Abduct.hs
```diff
@@ -0,0 +1,158 @@
...
+  
+  -- First, declare fresh constants for each of the three variables x, y, z.
+  x <- freshConstant sym (safeSymbol "x") BaseIntegerRepr
+  y <- freshConstant sym (safeSymbol "y") BaseIntegerRepr
+  z <- freshConstant sym (safeSymbol "z") BaseIntegerRepr
+
+  -- Next, build up the clause
+  zero <- intLit sym 0                        -- 0
+  pxyz <- intAdd sym x =<< intAdd sym y z -- x + y + z
+  ygte0 <- intLe sym zero y               -- 0 <= y
```
RyanGlScott, line 141: Minor formatting suggestion:

```suggestion
  zero <- intLit sym 0                    -- 0
  pxyz <- intAdd sym x =<< intAdd sym y z -- x + y + z
  ygte0 <- intLe sym zero y               -- 0 <= y
```

what4/test/Abduct.hs
```diff
@@ -0,0 +1,158 @@
...
+    [ -- test passes if f is disproved (~f is sat)
+      testSatAbd sym f [ ("x", x)
+                     , ("y", y)
+                     , ("z", z)
+                           ],
+      -- test passes if cvc5 returns 3 abducts (offline)
+      testAbdOffline sym f [ ("x", x)
+                     , ("y", y)
+                     , ("z", z)
+                     ],
```
RyanGlScott, line 155: Minor formatting suggestion:

```suggestion
      testSatAbd sym f [ ("x", x)
                       , ("y", y)
                       , ("z", z)
                       ],
      -- test passes if cvc5 returns 3 abducts (offline)
      testAbdOffline sym f [ ("x", x)
                           , ("y", y)
                           , ("z", z)
                           ],
```

RyanGlScott (commented):
Thanks for adding a test case! I'm quite happy with how it looks, modulo some minor suggestions.

what4/test/Abduct.hs
arjunvish: I used it because of a name clash with `Data.Text.length`

RyanGlScott: Ah, I overlooked that. Indeed, it's generally not advisable to import `Data.Text` unqualified, as its API overlaps with that of the `Prelude`. I'd recommend doing this instead:

```
import qualified Data.Text as Text
```

And then qualify all uses of `Text` functions (e.g., `Text.length`). That way, you shouldn't need to import `Data.List` at all.

arjunvish:
Fixes #214

RyanGlScott (approved):
Excellent work, @arjunvish. Thanks!

PR #216, opened 2022-08-29
Use more abstract domain information in intMax/Min

### LINKED ISSUE #215: `intMax` is sensitive to argument ordering when applying abstract domains
As an example, for some integer `i` with an abstract domain of `0 <= i`, we would expect `intMax 0 i = i` and `intMax i 0 = i` to always be the case. However at the moment we get that `intMax i 0 = (if i <= 0 then 0 else i)`.

This is a problem for the default constructor for `SymNat`, since it is almost never able to reduce the incoming integer based on its abstract domain.

A short term fix for `SymNat` would be to flip the order that the arguments are applied in, but it seems like this is a more general problem for any usage of `intMax`.

Somewhat related to #101

danmatichuk:
For intMax/Min x y, checks if either x <= y or y <= x
can be statically determined.
Explicitly sets the abstract domain of the result
to be rangeMax/Min of the arguments.

Fixes #215.

### DIFF
--- a/what4/src/What4/Interface.hs
+++ b/what4/src/What4/Interface.hs
@@ -740,14 +740,42 @@ class ( IsExpr (SymExpr sym), HashableF (SymExpr sym)
   -- | Return the minimum value of two integers.
   intMin :: sym -> SymInteger sym -> SymInteger sym -> IO (SymInteger sym)
   intMin sym x y =
-    do p <- intLe sym x y
-       intIte sym p x y
+    do x_le_y <- intLe sym x y
+       y_le_x <- intLe sym y x
+       case (asConstantPred x_le_y, asConstantPred y_le_x) of
+         -- x <= y
+         (Just True, _) -> return x
+         -- x < y
+         (_, Just False) -> return x
+         -- y < x
+         (Just False, _) -> return y
+         -- y <= x
+         (_, Just True) -> return y
+         _ ->
+           do let rng_x = integerBounds x
+              let rng_y = integerBounds y
+              unsafeSetAbstractValue (rangeMin rng_x rng_y) <$>
+                intIte sym x_le_y x y
 
   -- | Return the maximum value of two integers.
   intMax :: sym -> SymInteger sym -> SymInteger sym -> IO (SymInteger sym)
   intMax sym x y =
-    do p <- intLe sym x y
-       intIte sym p y x
+    do x_le_y <- intLe sym x y
+       y_le_x <- intLe sym y x
+       case (asConstantPred x_le_y, asConstantPred y_le_x) of
+         -- x <= y
+         (Just True, _) -> return y
+         -- x < y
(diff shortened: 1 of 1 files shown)


RyanGlScott:
The approach here looks reasonable. Do you have a small test case that you could add which demonstrates this more precise abstract domain information being used to reduce a complex expression?

danmatichuk:
The actual target of this change is to make the simplification rules fire when adding the positivity clamp in `SymNat` (i.e. ensuring that `0 <= i --> intMax i 0 = i`). I can put some tests in the `ExprsTest` module that demonstrates generally what kind of simplifications should now be firing.

what4/test/ExprsTest.hs
```diff
@@ -91,8 +91,66 @@ testInt = testGroup "int operators"
...
+      liftIO $ withTestSolver $ \sym -> do 
+        j <- freshBoundedInt sym (safeSymbol "j") Nothing (Just c)
+        i <- freshBoundedInt sym (safeSymbol "i") (Just c) Nothing
+        max_j_i <- intMax sym j i
+        res1 <- intEq sym max_j_i i
+        asConstantPred res1 @=? Just True
+        max_i_j <- intMax sym i j
+        res2 <- intEq sym max_i_j i
+        asConstantPred res2 @=? Just True
+  , testProperty "lo_i <= i; lo_j <= j; -> (max lo_j lo_j) <= intMax i j" $
```
RyanGlScott, line 111: A minor thing: should `lo_j <= j; -> ...` be `lo_j <= j -> ...`? I'm not sure how to parse `; ->`.

what4/test/ExprsTest.hs
```diff
@@ -91,8 +91,66 @@ testInt = testGroup "int operators"
...
+      liftIO $ withTestSolver $ \sym -> do
+        j <- freshBoundedInt sym (safeSymbol "j") (Just c) Nothing
+        i <- freshBoundedInt sym (safeSymbol "i") Nothing (Just c)
+        min_j_i <- intMin sym j i
+        res1 <- intEq sym min_j_i i
+        asConstantPred res1 @=? Just True
+        min_i_j <- intMin sym i j
+        res2 <- intEq sym min_i_j i
+        asConstantPred res2 @=? Just True
+  , testProperty "i <= hi_i; j <= hi_j; -> intMin i j <= (min hi_j hi_j)" $
```
RyanGlScott, line 137: Should `j <= hi_j; -> ...` be `j <= hi_j -> ...`?

RyanGlScott (approved):

what4/test/ExprsTest.hs
danmatichuk: Good point, I'll just change it to just use `&&` so it looks more standard.

RyanGlScott (approved):

PR #226, opened 2022-12-13
SyGuS

andreistefanescu:
Add support for Syntax Guided Synthesis (SyGuS) and Constrained Horn Clauses (CHC).

### DIFF
--- a/what4/src/What4/Interface.hs
+++ b/what4/src/What4/Interface.hs
@@ -2729,6 +2729,10 @@ type family SymFn sym :: Ctx BaseType -> BaseType -> Type
 
 data SomeSymFn sym = forall args ret . SomeSymFn (SymFn sym args ret)
 
+-- | Wrapper for `SymFn` that concatenates the arguments and the return types.
+--
+-- This is useful for implementing `TestEquality` and `OrdF` instances for
+-- `SymFn`, and for using `SymFn` as a key or a value in a `MapF`.
 data SymFnWrapper sym ctx where
   SymFnWrapper :: forall sym args ret . SymFn sym args ret -> SymFnWrapper sym (args ::> ret)
 
@@ -2746,8 +2750,17 @@ class IsSymFn (fn :: Ctx BaseType -> BaseType -> Type) where
   -- | Get the return type of a function.
   fnReturnType :: fn args ret -> BaseTypeRepr ret
 
+  -- | Test whether two functions are equal.
+  --
+  -- The implementation may be incomplete, that is, if it returns `Just` then
+  -- the functions are equal, while if it returns `Nothing` then the functions
+  -- may or may not be equal. The result of `freshTotalUninterpFn` or
+  -- `definedFn` tests equal with itself.
   fnTestEquality :: fn args1 ret1 -> fn args2 ret2 -> Maybe ((args1 ::> ret1) :~: (args2 ::> ret2))
 
+  -- | Compare two functions for ordering.
+  --
+  -- The underlying equality test is provided by `fnTestEquality`.
   fnCompare :: fn args1 ret1 -> fn args2 ret2 -> OrderingF (args1 ::> ret1) (args2 ::> ret2)
 
 
@@ -2794,7 +2807,6 @@ class ( IsExprBuilder sym
       , IsSymFn (SymFn sym)
       , OrdF (SymExpr sym)
       , OrdF (BoundVar sym)
-      , OrdF (SymFnWrapper sym)
       ) => IsSymExprBuilder sym where
 
   ----------------------------------------------------------------------
--- a/what4/src/What4/Interface.hs
+++ b/what4/src/What4/Interface.hs
@@ -38,6 +38,12 @@ provide several type family definitions and class instances for @sym@:
 
   [@instance 'HashableF' ('SymExpr' sym)@]
 
+  [@instance 'OrdF' ('BoundVar' sym)@]
+
+  [@instance 'TestEquality' ('BoundVar' sym)@]
+
+  [@instance 'HashableF' ('BoundVar' sym)@]
+
 The canonical implementation of these interface classes is found in "What4.Expr.Builder".
 -}
 {-# LANGUAGE CPP #-}
@@ -72,6 +78,8 @@ module What4.Interface
     -- ** Expression recognizers
   , IsExpr(..)
   , IsSymFn(..)
+  , SomeSymFn(..)
+  , SymFnWrapper(..)
   , UnfoldPolicy(..)
   , shouldUnfold
 
@@ -202,6 +210,7 @@ import           Data.Parameterized.Classes
 import qualified Data.Parameterized.Context as Ctx
 import           Data.Parameterized.Ctx
 import           Data.Parameterized.Utils.Endian (Endian(..))
+import           Data.Parameterized.Map (MapF)
 import           Data.Parameterized.NatRepr
 import           Data.Parameterized.TraversableFC
 import qualified Data.Parameterized.Vector as Vector
@@ -587,7 +596,7 @@ instance (HashableF (SymExpr sym), TestEquality (SymExpr sym)) => Hashable (SymN
 -- of an undefined function is _not_ guaranteed to be equivalant to a free
 -- constant, and no guarantees are made about what properties such values
 -- will satisfy.
-class ( IsExpr (SymExpr sym), HashableF (SymExpr sym)
+class ( IsExpr (SymExpr sym), HashableF (SymExpr sym), HashableF (BoundVar sym)
       , TestEquality (SymAnnotation sym), OrdF (SymAnnotation sym)
       , HashableF (SymAnnotation sym)
       ) => IsExprBuilder sym where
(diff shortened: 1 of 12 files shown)


andreistefanescu:
@RyanGlScott do you know why the CI fails?

what4/src/What4/Protocol/SMTLib2/Syntax.hs
```diff
@@ -836,6 +846,29 @@ getAbduct nm p = Cmd $ "(get-abduct " <> Builder.fromText nm <> " " <> renderTer
...
+-- | Declare a SyGus function to synthesize with the given name, arguments, and
+-- return type.
+synthFun :: Text -> [(Text, Sort)] -> Sort -> Command
+synthFun f args ret_tp = Cmd $ app "synth-fun"
+  [ Builder.fromText f
+  , builder_list $ map (\(var, tp) -> app (Builder.fromText var) [unSort tp]) args
+  , unSort ret_tp
+  ]
+
+-- | Declare a SyGus variale with the given name and type.
```
langston-barrett, line 858:
```suggestion
-- | Declare a SyGus variable with the given name and type.
```

what4/src/What4/Interface.hs
```diff
@@ -2769,6 +2793,8 @@ instance Show InvalidRange where
 class ( IsExprBuilder sym
       , IsSymFn (SymFn sym)
       , OrdF (SymExpr sym)
+      , OrdF (BoundVar sym)
+      , OrdF (SymFnWrapper sym)
```
RyanGlScott, line 2797: I believe this constraint is redundant, as there is already a top-level `OrdF (SymFnWrapper sym)` instance.

what4/src/What4/Protocol/SMTLib2/Response.hs
```diff
@@ -71,6 +71,8 @@ data SMTResponse = AckSuccess
                  | AckSat
                  | AckUnsat
                  | AckUnknown
+                 | AckInfeasible
+                 | AckFail
```
RyanGlScott, line 75: These can only arise from SyGuS sessions, correct? Worth stating in the comments if so.

what4/src/What4/Protocol/SMTLib2/Syntax.hs
```diff
@@ -51,11 +51,17 @@ module What4.Protocol.SMTLib2.Syntax
   , getUnsatCore
   , getAbduct
   , getAbductNext
+    -- * SyGus
```
RyanGlScott, line 54: Nit pick: this PR is inconsistent about referring to "SyGuS" (with two capitalized S's) versus "SyGus" (with only one capitalized S). I _think_ the former is the more common convention, so we should use that consistently, such as in the line above. (There are other occurrences of "SyGus" in the PR to which this comment also applies.)

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -931,6 +947,19 @@ class (SupportTermOps (Term h)) => SMTWriter h where
...
+                  -> [(Text, Some TypeMap)]
+                  -> TypeMap tp
+                  -> Command h
+
+  declareVarCommand :: f h
+                    -> Text
+                    -> TypeMap tp
+                    -> Command h
+
+  constraintCommand :: f h -> Term h -> Command h
```
RyanGlScott, line 961: These functions deserve some Haddocks, as downstream libraries will need to know how to implement these methods for their own `SMTWriter` instances.

what4/test/InvariantSynthesis.hs
```diff
@@ -0,0 +1,129 @@
...
+  conj2 <- andPred sym inv_i_n not_lt_add_i_1_n
+  eq_add_i_1_n <- intEq sym add_i_1 n
+  impl3 <- impliesPred sym conj2 eq_add_i_1_n
+  impl4 <- notPred sym impl3
+
+  CVC5.runCVC5InOverride sym logData [impl4] $ \res -> isSat res @? "sat"
+
+  subst <- CVC5.runCVC5SyGuS sym logData [SomeSymFn inv] [impl0, impl1, impl2] >>= \case
+    Sat res -> return res
+    Unsat{} -> fail $ "Infeasible"
```
RyanGlScott, line 68: Nit pick: redundant `$`:

```suggestion
    Unsat{} -> fail "Infeasible"
```

There are some other occurrences of this pattern elsewhere in the file as well.

what4/src/What4/Solver/Z3.hs
```diff
@@ -235,3 +243,83 @@ instance OnlineSolver (SMT2.Writer Z3) where
       timeout feat (Just z3StrictParsing) mbIOh sym
 
   shutdownSolverProcess = SMT2.shutdownSolver Z3
+
+runZ3Horn ::
```
RyanGlScott, line 247: I'd appreciate some more Haddocks about the `Z3Horn` family of functions here. In particular:

* How do the `Horn` functions differ from the normal Z3 functions?
* What is the difference between the Horn functions here in the SyGuS CVC5 functions in `What4.Solver.CVC5`? Is one more powerful than the other?

what4/src/What4/Solver/CVC5.hs
```diff
@@ -244,3 +250,124 @@ instance OnlineSolver (SMT2.Writer CVC5) where
           timeout feat (Just cvc5StrictParsing) mbIOh sym
 
   shutdownSolverProcess = SMT2.shutdownSolver CVC5
+
+
+data CVC5_SyGuS = CVC5_SyGuS deriving Show
```
RyanGlScott, line 255: As far as I can tell, the only difference between `CVC5_SyGuS` and `CVC5` is the set of command-line flags that get passed to the `defaultSolverArgs` in their `SMTLib2GenericSolver` instances. Is this correct? If so, this would be worth documenting.

what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -884,6 +919,339 @@ parseBvArraySolverValue w v (SApp ["store", arr, idx, val]) = do
...
+-- | Given an expression, monadically either returns proof that it is a
+-- bitvector or throws an error.
+getBVProof :: (I.IsExpr ex, MonadError String m) => ex tp -> m (BVProof tp)
+getBVProof expr = case I.exprType expr of
+  BaseBVRepr n -> return $ BVProof n
+  t -> throwError $ "expected BV, found " ++ show t
+
+-- | Operator type descriptions for parsing s-expression of
+-- the form @(operator operands ...)@.
+data Op sym where
```
RyanGlScott, line 1035: This looks incredibly similar to [`Op`](https://github.com/GaloisInc/what4/blob/17483418396afa8384de4089cdfaab9e5bbf912d/what4/src/What4/Serialize/Parser.hs#L271-L332) in `What4.Serialize.Parser`, to the point where I almost wonder if this code was copy-pasted. Why not just import `What4.Serialize.Parser` instead?

what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -884,6 +919,339 @@ parseBvArraySolverValue w v (SApp ["store", arr, idx, val]) = do
...
+    , Just LeqProof <- testLeq (knownNat @1) w ->
+      liftIO $ Some <$> I.bvLit sym w bv
+  SAtom nm -> do
+    env <- asks procLetEnv
+    case HashMap.lookup nm env of
+      Just expr -> return $ expr
+      Nothing -> throwError ""
+  SApp ["let", SApp bindings_sexp, body_sexp] -> do
+    let_env <- HashMap.fromList <$> mapM
+      (\(SApp [SAtom nm, expr_sexp]) -> do
```
RyanGlScott, line 1137: Nit pick: this will trigger `-Wincomplete-uni-patterns` on GHC 9.2. We should provide a fall-through case as well to avoid warnings.

RyanGlScott (commented):
> do you know why the CI fails?

I think this is because the CI is using an especially old version of Nix. I've submitted #227 with a fix.

what4/src/What4/Protocol/SMTLib2.hs
andreistefanescu: this is copy-pasta code, with some SMT-LIB specific changes. I initially wanted to use `What4.Serialize.Parser`, but that's not exactly SMT-LIB, so the code there requires some work. I plan to merge this for now, to limit the scope of this change. todo https://github.com/GaloisInc/what4/issues/228

RyanGlScott: OK. At a minimum, can you include a comment somewhere indicating:

1. That this is cargo-culted from `What4.Serialize.Parser`, and
2. A summary of the changes between it and the code in `What4.Serialize.Parser`

what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -884,6 +919,344 @@ parseBvArraySolverValue w v (SApp ["store", arr, idx, val]) = do
...
+        Just Refl -> liftIO (Some <$> I.baseTypeIte sym arg1_expr arg2_expr arg3_expr)
+        Nothing -> throwError ""
+      _ -> throwError ""
+  SApp ["concat", arg1, arg2] -> do
+    Some arg1_expr <- parseExpr sym arg1
+    Some arg2_expr <- parseExpr sym arg2
+    BVProof{} <- getBVProof arg1_expr
+    BVProof{} <- getBVProof arg2_expr
+    liftIO $ Some <$> I.bvConcat sym arg1_expr arg2_expr
+--  SApp ["_", "extract", arg1, arg2, arg3] -> do
```
chameco, line 1169: Stray comment

what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -884,6 +919,344 @@ parseBvArraySolverValue w v (SApp ["store", arr, idx, val]) = do
...
+      case testEquality m n of
+        Just Refl -> liftIO (Some <$> op sym arg1 arg2)
+        Nothing -> throwError $ printf "arguments to %s must be the same length, \
+                                       \but arg 1 has length %s \
+                                       \and arg 2 has length %s"
+                                       operator
+                                       (show m)
+                                       (show n)
+    -- Nothing -> throwError ""
+    _ -> throwError ""
```
chameco, line 1211: Is it possible to throw a more descriptive error here / at other instances of `throwError ""`? Also, I'd prefer to explicitly match `Nothing` vs. `_` if we're handling all cases.

chameco (approved):
Since this is a large change, it would be nice to document some things in the PR description / eventual squashed commit message. In particular:
- broadly, what has changed that could affect existing codepaths (such as using Z3 normally through What4 from SAW) in case something breaks elsewhere - I suspect this isn't too much, but e.g. things like the `IsSymFn` typeclass change
- what new features were introduced, and what is the interface to those features

I think overall this looks good to merge, pending some additional documentation. Thanks for the tests for the new functionality. It might also be nice to use some of our existing proofs to make sure that existing functionality doesn't break - in the absence of Fryingpan, CI passing on a draft SAW PR that points to this branch would be reassuring.

what4/src/What4/Protocol/SMTLib2.hs
andreistefanescu: the plan is to merge back this code in the `What4.Serialize.Parser`, so I won't write error messages now.

PR #227, opened 2022-12-13
CI: Use Nix 2.11.0

RyanGlScott:
Nix 2.4 does not work well with GitHub Actions' Ubuntu 22.04 runners, as seen in https://github.com/cachix/install-nix-action/issues/141.

-----

I'm hoping that this addresses the CI woes encountered in #226.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,8 +14,8 @@ on:
 # * GHC versions
 #   - specified in the strategy matrix
 #   - provided via nix: ensure the nix base and NIXPKGS used provide the requested GHC version
-# * nix tool version 2.4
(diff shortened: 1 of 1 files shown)


kquick (approved):
Pinning the nix version is an intent to support repeatability into the future, but since that's not the only moving part, that promise isn't supportable. An alternative at this point would be to have the CI simply install the latest and perform `nix --version` into the logs; we would then have that information available in the logs for diagnosing future issues.  However, that might encounter the same issue in reverse: the latest `nix` version might become incompatible with the locked `ubuntu` version, so there's no silver bullet here.

I'm good with these changes (thanks!) but wanted to point out the alternative approaches for future consideration.

Issue #224, opened 2022-11-28
Make it build with ghc 9.4.3


yav:
One problem at the moment is that we depend on `s-cargot` which has a restrictive dependency on `text`.

I file a ticket downstream:
https://github.com/aisamanra/s-cargot/issues/15

RyanGlScott:
Other dependencies that will need to be updated:

* [ ] `parameterized-utils`
* [ ] `bv-sized`
* [ ] `bitwise`

PR #230, opened 2023-01-18
Support building with GHC 9.4

RyanGlScott:
This contains a varity of changes needed to make the libraries in the `what4` repo compile with GHC 9.4:

* GHC 9.4 is pickier about undecidable instance checking. See [this section](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.4?version_id=24540b698481cf2b0bd11029b58eaddc0fbfbb31#stricter-undecidableinstances-checking) of the GHC 9.4 Migration Guide. As a result, I had to explicitly enable `UndecidableInstances` in `What4.Utils.AbstractDomains` to make it compile.
* `ST` no longer has a `MonadFail` instance. See [this section](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.4?version_id=b60e52482a666d25638d59cd7e86851ddf971dc1#st-is-no-longer-an-instance-of-monadfail) of the GHC 9.4 Migration Guide. To adapt to this change, I removed the `MonadFail` instance for `VarRecorder`, which was built on top of `ST`'s `MonadFail` instance. Accordingly, all uses of `fail` at type `VarRecorder` have now been replaced with `error`.
* I bumped the `aig` submodule to bring in the changes from https://github.com/GaloisInc/aig/pull/13, which are required to make `aig` build with GHC 9.4.
* I bumped the upper version bounds of various dependencies to allow them to build with GHC 9.4.

Fixes https://github.com/GaloisInc/what4/issues/224.


kquick (approved):
As a minor suggestion, if you wanted to use 9.4.3 instead of 9.4.4, then the `nixpkgs/22.11` could be used instead of `nixpkgs-unstable` (with the advantage that the former is... more stable), but we can also address this via the upcoming 23.05 release as well, so it's your call.

RyanGlScott:
I don't have a strong opinion on which minor 9.4 release to use, so I've changed it to GHC 9.4.3 / `nixpkgs/22.05`.

kquick:
Ah, my mistake.  I was getting the results of a searchlist when I looked to see what GHC version 22.11 had and a more recent set of packages was also available on the searchlist path.  You **do** need to use `nixos-unstable` at present to get `9.4.4`... sorry about that!

PR #235, opened 2023-07-19
Support building with GHC 9.6

RyanGlScott:
This patch contains a variety of fixes needed to build the libraries in the
`what4` repo with GHC 9.6:

* GHC 9.6 bundles `mtl-2.3.*`, which no longer re-exports `Control.Monad`,
  `Control.Monad.IO.Class`, and similar modules from `mtl`-related modules.
  To accommodate this, various imports have been made more explicit.
* `MonadTrans t` now has a quantified `forall m. Monad m => Monad (t m)`
  superclass in `mtl-2.3.*`. As a result, the `MonadTrans (PartialT sym)`
  instance must now have an `IsExpr (SymExpr sym)` instance context in order
  for it to typecheck, as this is the same instance context that `PartialT`'s
  `Monad` instance has.

  This is technically a breaking change, so I have noted it in the `what4`
  changelog. That being said, this change is unlikely to affect many people
  in practice, considering that the `MonadTrans` instance for `PartialT` is
  usually used in tandem with the `Monad` instance.
* Various upper version bounds on `base` have been lifted to allow building
  with `base-4.18`.
* The `aig` submodule has been bumped to bring in the changes from
  https://github.com/GaloisInc/aig/pull/15, which allows it to build with GHC 9.6.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -104,6 +104,7 @@ jobs:
             9.0.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-22.05 ;;
             9.2.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-22.05 ;;
             9.4.4) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-unstable ;;
+            9.6.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-unstable ;;
             *)     GHC_NIXPKGS=github:nixos/nixpkgs/21.11 ;;
           esac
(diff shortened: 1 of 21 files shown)


.github/workflows/test.yml
```diff
@@ -104,6 +104,7 @@ jobs:
             9.0.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-22.05 ;;
             9.2.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-22.05 ;;
             9.4.4) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-unstable ;;
+            9.6.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-unstable ;;
```
kquick, line 107: Both 9.6.2 and 9.4.4 can be obtained from `nixos-23.05` instead of `nixos-unstable`.  Since the latter is a moving target, we should use the former for these now that it is available.

kquick (approved):

PR #245, opened 2023-12-05
what4: Don't annotate bound variables

### LINKED ISSUE #244: Do bound variables need annotations?
The annotation mechanism (`annotateTerm`) attaches `Nonce`s to expressions via the `Annotation` constructor of `NonceApp`. [Bound variables](https://github.com/GaloisInc/what4/blob/076f49b996bd328764a9c6edece6b7c34cb1cfb4/what4/src/What4/Expr/App.hs#L664) already have associated `Nonce`s, their `bvarId`s. We should consider whether `annotateTerm`, when applied to a `BoundExpr`, should return the term as-is, and `getAnnotation` could just return the `bvarId`.

langston-barrett:
Fixes #244. Should result in slightly better performance for code that annotates bound variables, and prettier printing.

### DIFF
--- a/what4/src/What4/Expr/Builder.hs
+++ b/what4/src/What4/Expr/Builder.hs
@@ -1630,6 +1630,7 @@ instance IsExprBuilder (ExprBuilder t st fs) where
 
   annotateTerm sym e =
     case e of
+      BoundVarExpr (bvarId -> n) -> return (n, e)
       NonceAppExpr (nonceExprApp -> Annotation _ n _) -> return (n, e)
       _ -> do
         let tpr = exprType e
@@ -1639,11 +1640,13 @@ instance IsExprBuilder (ExprBuilder t st fs) where
 
   getAnnotation _sym e =
     case e of
+      BoundVarExpr (bvarId -> n) -> Just n
       NonceAppExpr (nonceExprApp -> Annotation _ n _) -> Just n
       _ -> Nothing
 
   getUnannotatedTerm _sym e =
     case e of
+      BoundVarExpr {} -> Just e
       NonceAppExpr (nonceExprApp -> Annotation _ _ x) -> Just x
       _ -> Nothing
 
--- a/what4/src/What4/Expr/Builder.hs
+++ b/what4/src/What4/Expr/Builder.hs
@@ -1646,7 +1646,6 @@ instance IsExprBuilder (ExprBuilder t st fs) where
 
   getUnannotatedTerm _sym e =
     case e of
-      BoundVarExpr {} -> Just e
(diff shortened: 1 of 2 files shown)


what4/src/What4/Expr/Builder.hs
```diff
@@ -1639,6 +1640,7 @@ instance IsExprBuilder (ExprBuilder t st fs) where
 
   getAnnotation _sym e =
     case e of
+      BoundVarExpr (bvarId -> n) -> Just n
       NonceAppExpr (nonceExprApp -> Annotation _ n _) -> Just n
       _ -> Nothing
```
langston-barrett, line 1646: Note that I did *not* change `getUnannotatedTerm` to return `Just` for `BoundVarExpr`s. Let me know if anyone thinks I should.

RyanGlScott: I'm inclined to make the implementations of `annotateTerm` and `getUnannotatedTerm` also special-case `BoundVarExpr`, as doing so feels like it would avoid some potential surprising behavior down the line.

langston-barrett: Done! Let me know if the implementation looks unsurprising.

langston-barrett:
FWIW, I've tested this against a (unfortunately, closed-source) codebase that makes heavy use of the annotation mechanism and these changes don't seem to affect it.

what4/src/What4/Expr/Builder.hs
```diff
@@ -1639,11 +1640,13 @@ instance IsExprBuilder (ExprBuilder t st fs) where
...
 
   getUnannotatedTerm _sym e =
     case e of
+      BoundVarExpr {} -> Just e
```
RyanGlScott, line 1649: This is a bit of a strange thing to do, since the expression that we returning here _is_ annotated (with the same annotation as the argument expression, no less). I suppose I imagined that `BoundVarExpr` had an underlying `Expr` to return, but upon closer inspection, it doesn't. Hm.

In light of this, perhaps `getUnannotatedTerm` should return `Nothing` for `BoundVarExpr`. We should also amend the Haddocks for `getUnannotatedTerm` to mention that returning `Nothing` doesn't imply that the argument expression is completely unannotated, but rather that it is not feasible to separate the expression from the annotation. If we do this, then I would be less bothered by the fact that `getUnannotatedTerm` has fewer special cases than `annotateTerm`/`getAnnotation` do.

langston-barrett: Removed, documented!

RyanGlScott (approved):

PR #247, opened 2023-12-06
what4: Don't annotate `{Nonce,}AppExpr`s

### LINKED ISSUE #246: Do `{Nonce,}AppExpr`s need annotations?
This is a follow-up to #244, where we observed that `BoundVarExpr`s already carry `Nonce`s, so there's no need to create `Annotation` expressions for them. As it turns out, the same is true for `{Nonce,}AppExpr`:

https://github.com/GaloisInc/what4/blob/076f49b996bd328764a9c6edece6b7c34cb1cfb4/what4/src/What4/Expr/App.hs#L111

https://github.com/GaloisInc/what4/blob/076f49b996bd328764a9c6edece6b7c34cb1cfb4/what4/src/What4/Expr/App.hs#L125

Perhaps we should special-case these in the same way we did `BoundVarExpr` in #245.

langston-barrett:
Fixes #246. Really, `Annotation` shouldn't need to carry a `Nonce`, because the outer `NonceAppExpr` will already have one. However, actually removing it is challenging due to the return type of `sbNonceExpr`, and in turn `ExprAllocator`'s `nonceExpr`, which returns an `Expr`. Surely in practice, this is always a `NonceAppExpr`, but there's no way to tell from the type. To avoid a larger refactor or introducing partiality, I'm keeping the `Nonce` in `Annotation` for now.

Fixes #246, though we may want to create a follow-up about removing the `Nonce` from `Annotation`.

### DIFF
--- a/what4/src/What4/Expr/App.hs
+++ b/what4/src/What4/Expr/App.hs
@@ -678,9 +678,19 @@ data ExprBoundVar t (tp :: BaseType) =
 -- go. Uses of the 'NonceApp' type will tie the knot through this
 -- parameter. Parameter @tp@ indicates the type of the expression.
 data NonceApp t (e :: BaseType -> Type) (tp :: BaseType) where
+  -- | See 'What4.Interface.annotateTerm'.
   Annotation ::
     !(BaseTypeRepr tp) ->
     !(Nonce t tp) ->
+    !(e tp) ->
+    NonceApp t e tp
+
+  -- | See 'What4.Interface.opacify'.
+  --
+  -- No rewrite rules in 'IsExprBuilder' should examine the subterms of
+  -- 'Opaque', this is the contract of 'What4.Interface.opacify'.
+  Opaque ::
+    !(BaseTypeRepr tp) ->
     !(e tp) ->
     NonceApp t e tp
 
@@ -1778,6 +1788,7 @@ nonceAppType :: IsExpr e => NonceApp t e tp -> BaseTypeRepr tp
 nonceAppType a =
   case a of
     Annotation tpr _ _ -> tpr
+    Opaque tpr _ -> tpr
     Forall{} -> knownRepr
     Exists{} -> knownRepr
     ArrayFromFn   fn       -> BaseArrayRepr (symFnArgTypes fn) (symFnReturnType fn)
@@ -1917,6 +1928,7 @@ quantAbsEval :: IsExpr e =>
 quantAbsEval f q =
   case q of
     Annotation _ _ v -> f v
+    Opaque _ v -> f v
     Forall _ v -> f v
     Exists _ v -> f v
     ArrayFromFn _       -> unconstrainedAbsValue (nonceAppType q)
@@ -2305,6 +2317,7 @@ ppNonceApp :: forall m t e tp
 ppNonceApp ppFn a0 = do
   case a0 of
     Annotation _ n x -> pure $ prettyApp "annotation" [ showPrettyArg n, exprPrettyArg x ]
+    Opaque _ x -> pure $ prettyApp "opaque" [ exprPrettyArg x ]
     Forall v x -> pure $ prettyApp "forall" [ stringPrettyArg (ppBoundVar v), exprPrettyArg x ]
     Exists v x -> pure $ prettyApp "exists" [ stringPrettyArg (ppBoundVar v), exprPrettyArg x ]
     ArrayFromFn f -> resolve <$> ppFn f
--- a/what4/src/What4/Interface.hs
+++ b/what4/src/What4/Interface.hs
@@ -407,7 +407,7 @@ class HasAbsValue e => IsExpr e where
   -- Note that composing expressions together can sometimes widen the abstract
   -- domains involved, so if you use this function to change an abstract value,
   -- be careful than subsequent operations do not widen away the value. As a
-  -- potential safeguard, one can use 'annotateTerm' on the new expression to
+  -- potential safeguard, one can use 'opacify' on the new expression to
   -- inhibit transformations that could change the abstract value.
   unsafeSetAbstractValue :: AbstractValue tp -> e tp -> e tp
 
@@ -638,6 +638,9 @@ class ( IsExpr (SymExpr sym), HashableF (SymExpr sym), HashableF (BoundVar sym)
   -- | Set current location of program for term creation purposes.
   setCurrentProgramLoc :: sym -> ProgramLoc -> IO ()
 
+  ----------------------------------------------------------------------
+  -- Basic operations
+
   -- | Return true if two expressions are equal. The default
   -- implementation dispatches 'eqPred', 'bvEq', 'natEq', 'intEq',
   -- 'realEq', 'cplxEq', 'structEq', or 'arrayEq', depending on the
@@ -676,6 +679,9 @@ class ( IsExpr (SymExpr sym), HashableF (SymExpr sym), HashableF (BoundVar sym)
       BaseStructRepr{} -> structIte sym c x y
       BaseArrayRepr{}  -> arrayIte  sym c x y
 
+  ----------------------------------------------------------------------
+  -- Annotations and opaque expressions
+
   -- | Given a symbolic expression, annotate it with a unique identifier
   --   that can be used to maintain a connection with the given term.
   --   The 'SymAnnotation' is intended to be used as the key in a hash
@@ -699,6 +705,14 @@ class ( IsExpr (SymExpr sym), HashableF (SymExpr sym), HashableF (BoundVar sym)
   --   This returns 'Nothing' for terms that do not have annotations,
   --   or for terms that cannot be separated from their annotations.
   getUnannotatedTerm :: sym -> SymExpr sym tp -> Maybe (SymExpr sym tp)
+
+  -- | Make an expression /opaque/, inhibiting rewrites that, in many cases,
+  -- coarsen abstract domain information. See #248 for discussion and #249 for
+  -- progress towards obviating opaque expressions.
+  opacify :: sym -> SymExpr sym tp -> IO (SymExpr sym tp)
+
+  -- | Inverse operation of 'opacify'.
+  clarify :: sym -> SymExpr sym tp -> SymExpr sym tp
 
   ----------------------------------------------------------------------
   -- Boolean operations.
--- a/what4/CHANGES.md
+++ b/what4/CHANGES.md
@@ -4,6 +4,9 @@
   wrapper around `BoolMap` that makes clear that the `BoolMap` in question
   represents a conjunction (as `BoolMap`s may also represent disjunctions).
   See the Haddocks on `ConjMap` for more details.
+* Add `opacify`, `clarify` to `IsExprBuilder`, for preserving abstract values.
+  These should be used instead of `annotateTerm` for that purpose.
+* Modify `annotateTerm` to not annotate `{Nonce,}AppExpr`s.
 
 # 1.6.2 (Sep 2024)
 
--- a/what4/src/What4/Expr/AppTheory.hs
+++ b/what4/src/What4/Expr/AppTheory.hs
@@ -42,6 +42,7 @@ quantTheory :: NonceApp t (Expr t) tp -> AppTheory
 quantTheory a0 =
   case a0 of
     Annotation tpr _ _ -> typeTheory tpr
+    Opaque tpr _ -> typeTheory tpr
     Forall{} -> QuantifierTheory
     Exists{} -> QuantifierTheory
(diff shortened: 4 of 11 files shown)


langston-barrett:
Printf-debugging shows that this branch hits this case in [`semiRingAdd`](https://github.com/GaloisInc/what4/blob/e46dff42096550d860d210bd9cf6e435d8cd7ce5/what4/src/What4/Expr/Builder.hs#L1391):

https://github.com/GaloisInc/what4/blob/e46dff42096550d860d210bd9cf6e435d8cd7ce5/what4/src/What4/Expr/Builder.hs#L1401

whereas `master` hits the base case, likely because the "annotated" expression is (incorrectly) not recognized as a semiring sum:

https://github.com/GaloisInc/what4/blob/e46dff42096550d860d210bd9cf6e435d8cd7ce5/what4/src/What4/Expr/Builder.hs#L1425

Counter-intuitively, it appears that hitting the more specialized case results in a less-specific answer. Possibly, the `WSum.addConstant` code doesn't take abstract domains into account as carefully as `semiRingSum`?

langston-barrett:
I think this may have to do with the fact that the catch-all case calls this function:

https://github.com/GaloisInc/what4/blob/e46dff42096550d860d210bd9cf6e435d8cd7ce5/what4/src/What4/Expr/Builder.hs#L1213-L1220

whereas the more specific case calls `sum'` directly, bypassing `WSum.asConstant`. There are two possible solutions:

1. Call `semiRingSum` ranther than `sum'` in the `(SR_Sum, SR_Constant)` case. This could be beneficial to performance if this case is often reducible to a constant, or detrimental if it's not. I'm not sure why `sum'` is called directly, perhaps it's the assumption that it's not common.
2. Call `WSum.asConstant` in `asBV`. Again, this could possibly make a lot of other code faster or slower depending on how common the case of a constant-able sum appears.

langston-barrett:
It appears that `WSum.addVars` (in `master`/the base case) is hitting this case of `sbMakeExpr`

https://github.com/GaloisInc/what4/blob/e46dff42096550d860d210bd9cf6e435d8cd7ce5/what4/src/What4/Expr/Builder.hs#L589

whereas the result of `WSum.addConstant` (on this branch) hits the catch-all case:

https://github.com/GaloisInc/what4/blob/e46dff42096550d860d210bd9cf6e435d8cd7ce5/what4/src/What4/Expr/Builder.hs#L590

So I need to figure out why `abstractEval` treats these differently.

langston-barrett:
Okay, I've managed to whittle down the issue to a test case, reported here (as it happens on `master`, not just on this branch): https://github.com/GaloisInc/what4/issues/248

langston-barrett:
Alright, I've worked around the issues with the abstract domains by introducing a new type of `NonceExpr`, namely `Opaque`, which is specifically designed to preserve abstract values. See the docs and CHANGELOG.

langston-barrett:
I'm trying this out on a codebase that makes extensive use of annotations.

what4/src/What4/Expr/App.hs
```diff
@@ -678,12 +678,22 @@ data ExprBoundVar t (tp :: BaseType) =
...
     NonceApp t e tp
 
+  -- | See 'What4.Interface.opacify'.
+  --
+  -- No rewrite rules in 'IsExprBuilder' should examine the subterms of
+  -- 'Opaque', this is the contract of 'What4.Interface.opacify'.
+  Opaque ::
+    !(BaseTypeRepr tp) ->
+    !(e tp) ->
+    NonceApp t e tp
```
RyanGlScott, line 695: What part of the code ignores the subterms of `Opaque`? (It's not obvious to me at a glance where the magic actually happens.)

what4/src/What4/Interface.hs
```diff
@@ -407,7 +407,7 @@ class HasAbsValue e => IsExpr e where
   -- Note that composing expressions together can sometimes widen the abstract
   -- domains involved, so if you use this function to change an abstract value,
   -- be careful than subsequent operations do not widen away the value. As a
```
RyanGlScott, line 409:
```suggestion
  -- be careful that subsequent operations do not widen away the value. As a
```

what4/src/What4/Interface.hs
```diff
@@ -700,6 +706,14 @@ class ( IsExpr (SymExpr sym), HashableF (SymExpr sym), HashableF (BoundVar sym)
   --   or for terms that cannot be separated from their annotations.
   getUnannotatedTerm :: sym -> SymExpr sym tp -> Maybe (SymExpr sym tp)
 
+  -- | Make an expression /opaque/, inhibiting rewrites that, in many cases,
+  -- coarsen abstract domain information. See #248 for discussion and #249 for
+  -- progress towards obviating opaque expressions.
+  opacify :: sym -> SymExpr sym tp -> IO (SymExpr sym tp)
```
RyanGlScott, line 712: After reading these Haddocks, I'm a bit confused about the relationship between `opacify` and `annotateTerm`. Is the idea that `annotateTerm` should make no guarantees about inhibiting rewrites, as that should be the role of `opacify`? Or is the idea that `annotateTerm` _should_ guarantee that rewrites are inhibited, and that the only reason it doesn't are due to bugs (e.g., #248)? If it's the latter, is `opacify` meant to be a short-term workaround until bugs like #248 are fixed?

what4/src/What4/Interface.hs
```diff
@@ -700,6 +706,14 @@ class ( IsExpr (SymExpr sym), HashableF (SymExpr sym), HashableF (BoundVar sym)
...
+
+  -- | Inverse operation of 'opacify'.
+  clarify :: sym -> SymExpr sym tp -> SymExpr sym tp
```
RyanGlScott, line 715: It would be helpful to specify what happens if you call `clarify` on a `SymExpr` that wasn't returned by a call to `opacify`.

RyanGlScott:
Do you expect that these changes to how annotations work will have effects on downstream code? If so, is it worth advertising this in the changelog?

what4/src/What4/Expr/App.hs
langston-barrett: Ah, it's just that the implementation of `IsExprBuilder` for `ExprBuilder` doesn't match on it and rewrite under it! That's the second part of the Haddock here.

what4/src/What4/Interface.hs
langston-barrett: > Is the idea that annotateTerm should make no guarantees about inhibiting rewrites, as that should be the role of opacify?

Exactly! `annotateTerm` was previously treated as if it made this guarantee, but I think it's better to have a separate constructor (`Opaque`) for that (which doesn't need to attach an extra `Nonce`, either).

> Or is the idea that annotateTerm should guarantee that rewrites are inhibited, and that the only reason it doesn't are due to bugs (e.g., https://github.com/GaloisInc/what4/issues/248)?

The test case in that issue *doesn't* use `annotateTerm`, which is *why* the abstract value is discarded. `annotateTerm` does happen to prevent rewrites at the moment, but for the sake of the changes in this PR, I think we should abandon that.

> If it's the latter, is opacify meant to be a short-term workaround until bugs like https://github.com/GaloisInc/what4/issues/248 are fixed?

Yes, it is a temporary workaround for that issue.

langston-barrett:
> Do you expect that these changes to how annotations work will have effects on downstream code? If so, is it worth advertising this in the changelog?

Yes, in that using `annotateTerm` to preserve abstract values won't work as well and code that wants to do that should use `opacify` instead - which is what I put in the changelog. I'm open to suggestions on wording to make this more clear!

what4/src/What4/Expr/App.hs
RyanGlScott: Sorry, I must be missing something—where exactly does this happen in the `IsExprBuilder` instance for `ExprBuilder`?

what4/src/What4/Interface.hs
RyanGlScott: > Yes, it is a temporary workaround for that issue.

I am confused. If `annotateTerm` is no longer meant to inhibit rewrites, then isn't `opacify` the only part of the API that serves this role? If so, I would expect `opacify` to be a stable part of the API, but your wording suggests that it isn't... I must be missing something.

RyanGlScott:
OK. I think I have large SAW proofs in mind when I wrote https://github.com/GaloisInc/what4/pull/247#issuecomment-2653907570 , as I worry that this could lead to subtle changes in proof goals that are hard to track down. If all that is required is to audit uses of `annotateTerm` and replace them with `opacify`, then perhaps this isn't too bad.

what4/src/What4/Expr/App.hs
langston-barrett: > Sorry, I must be missing something—where exactly does this happen in the IsExprBuilder instance for ExprBuilder?

That's the thing haha: It *doesn't* happen. There *are no pattern matches* on `Opaque`, which is what makes it preserve the abstract values. It works just like `Annotation` did for this purpose.

what4/src/What4/Interface.hs
langston-barrett: > If annotateTerm is no longer meant to inhibit rewrites, then isn't opacify the only part of the API that serves this role?

Yes, `opacify` is meant to *replace* the use of `annotateTerm` for this purpose, because the primary changes in this MR (not attaching `Annotation` to `{Nonce,}AppExpr`s) make `annotateTerm` no longer suitable for it (as demonstrated by the weird test case failures above).

> If so, I would expect opacify to be a stable part of the API, but your wording suggests that it isn't... I must be missing something.

Well, the entire use of `annotateTerm` to preserve abstract values was a hack to work around #248 and issues like it - If all the rewrite rules preserved abstract values, there would be no need to hide terms under `Annotation`/`Opaque` to preserve their abstract values. This behavior was not documented as part of the intent of the annotation API. So IMO, this is replacing one temporary workaround with one that is better-documented and more efficient.

what4/src/What4/Expr/App.hs
langston-barrett: To give an example of what I'm talking about: #248 happens because [`bvAdd`](https://github.com/GaloisInc/what4/blob/ebf9ac050c4ef5a5fda880e2e92cf76e14028ad2/what4/src/What4/Expr/Builder.hs#L3146) calls [`semiRingAdd`](https://github.com/GaloisInc/what4/blob/ebf9ac050c4ef5a5fda880e2e92cf76e14028ad2/what4/src/What4/Expr/Builder.hs#L1789), which performs some rewrites using pattern-matching like so:

https://github.com/GaloisInc/what4/blob/ebf9ac050c4ef5a5fda880e2e92cf76e14028ad2/what4/src/What4/Expr/Builder.hs#L1798

Specifically, that issue calls out this case as problematic:

https://github.com/GaloisInc/what4/blob/ebf9ac050c4ef5a5fda880e2e92cf76e14028ad2/what4/src/What4/Expr/Builder.hs#L1806

But note that none of those patterns would match if either `x` or `y` were `Opaque`, so the rewrite rules wouldn't fire, so the abstract domain would be preserved. This is how it works for `Annotation` on `master`.

what4/src/What4/Interface.hs
RyanGlScott: > Well, the entire use of `annotateTerm` to preserve abstract values was a hack to work around https://github.com/GaloisInc/what4/issues/248 and issues like it - If all the rewrite rules preserved abstract values, there would be no need to hide terms under `Annotation`/`Opaque` to preserve their abstract values.

Ah, you had preserving abstract values in mind when you wrote your comments—I see. I _think_ I agree that if everything preserved abstract values, then we'd no longer need `Opaque`, although it's hard to say for sure that there wouldn't be other use cases for `Opaque`. For instance, it's conceivable that there are SAW proof goals that are easier to prove when derived from certain `what4` `Expr` shapes than others, and `Opaque` could be one possible mechanism to achieve this. (I don't have a concrete example of this in mind, but it doesn't seem totally outside of the realm of possibility.)

In that case, I think if we added some additional clarification to the `annotateTerm` Haddocks about what it does (and does not) guarantee, then I'd be content with this.

what4/src/What4/Expr/App.hs
RyanGlScott: > But note that none of those patterns would match if either x or y were `Opaque`

I'm not sure how to typecheck this sentence, since `x` and `y` are `Expr`s, but `Opaque` is a `NonceApp` (as is `Annotation`). Surely `Opaque` wouldn't matter here?

RyanGlScott: Thank you! This is the part that I was missing.

langston-barrett:
I don't currently have the bandwidth to push this forward.

PR #250, opened 2024-01-10
BV <> LIA translation for CHC solvers 

andreistefanescu:


what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -1166,6 +1176,24 @@ parseExpr sym sexp = case sexp of
...
+        _ -> throwError ""
+  SApp ((SAtom operator) : operands)
+    | Just op <- HashMap.lookup operator (opTable @sym)
+    , Just assoc <- opAssoc op
+    , length operands > 2 -> case assoc of
+      LeftAssoc -> parseExpr sym $ foldl' (\acc arg -> SApp [SAtom operator, acc, arg]) (head operands) (tail operands)
+      RightAssoc -> parseExpr sym $ foldr' (\arg acc -> SApp [SAtom operator, arg, acc]) (last operands) (init operands)
   SApp ((SAtom operator) : operands) -> case HashMap.lookup operator (opTable @sym) of
     Just (Op1 arg_types fn) -> do
       args <- mapM (parseExpr sym) operands
```
RyanGlScott, line 1199: Both of these cases overlap to some degree. To make this code easier to follow, would you be OK with refactoring it to look more like this?

```hs
  SApp ((SAtom operator) : operands) -> case HashMap.lookup operator (opTable @sym) of
    Just op <- HashMap.lookup operator (opTable @sym)
      | Just assoc <- opAssoc op
      , length operands > 2 -> case assoc of
        LeftAssoc -> parseExpr sym $ foldl' (\acc arg -> SApp [SAtom operator, acc, arg]) (head operands) (tail operands)
        RightAssoc -> parseExpr sym $ foldr' (\arg acc -> SApp [SAtom operator, arg, acc]) (last operands) (init operands)
    Just (Op1 arg_types fn) -> do
      ...
```

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -831,9 +831,9 @@ cacheLookupFnNameBimap :: WriterConn t h -> [SomeExprSymFn t] -> IO (Bimap (Some
 cacheLookupFnNameBimap conn fns = Bimap.fromList <$> mapM
   (\some_fn@(SomeExprSymFn fn) -> do
     maybe_smt_sym_fn <- cacheLookupFn conn $ symFnId fn
-    case maybe_smt_sym_fn of
-      Just (SMTSymFn nm _ _) -> return (some_fn, nm)
-      Nothing -> fail $ "Could not find function in cache: " ++ show fn)
+    return $ case maybe_smt_sym_fn of
+      Just (SMTSymFn nm _ _) -> (some_fn, nm)
+      Nothing -> (some_fn, solverSymbolAsText $ symFnName fn))
```
RyanGlScott, line 836: Perhaps worth documenting the fact that this function never fails if a `SomeExprSymFn`'s name cannot be looked up, but it will instead map that `SomeExprSymFn` to itself.

what4/src/What4/Solver/Z3.hs
```diff
@@ -288,18 +302,54 @@ runZ3Horn sym log_data inv_fns horn_clauses = do
...
+                    , Just Refl <- testEquality (fnReturnType fn) (fnReturnType defined_fn) ->
+                      return $ MapF.Pair (SymFnWrapper fn) (SymFnWrapper defined_fn)
+                  _ -> fail $ "runZ3Horn: unexpected function " ++ show fn
+              | otherwise -> fail $ "runZ3Horn: unexpected function " ++ show fn)
+          inv_fns
+
+        all_unsat <- and <$> mapM
+          (\clause -> do
+            defined_clause <- notPred sym =<< substituteSymFns sym defined_inv_fns clause
+            runZ3InOverride sym (log_data { logVerbosity = 2 }) [defined_clause] $ return . isUnsat)
```
RyanGlScott, line 329: Why does this use `logVerbosity = 2` instead of inheriting the verbosity from the provided `log_data`?

what4/src/What4/Solver/Z3.hs
```diff
@@ -288,18 +302,54 @@ runZ3Horn sym log_data inv_fns horn_clauses = do
...
+            if| Just (SomeSymFn lia_fn) <- Map.lookup (SomeSymFn fn) bv_to_lia_fn_subst
+              , Just (SymFnWrapper lia_defined_fn) <- MapF.lookup (SymFnWrapper lia_fn) lia_defined_fns -> do
+                some_defined_fn <- transformSymFnLIA2BV sym $ SomeSymFn lia_defined_fn
+                case some_defined_fn of
+                  SomeSymFn defined_fn
+                    | Just Refl <- testEquality (fnArgTypes fn) (fnArgTypes defined_fn)
+                    , Just Refl <- testEquality (fnReturnType fn) (fnReturnType defined_fn) ->
+                      return $ MapF.Pair (SymFnWrapper fn) (SymFnWrapper defined_fn)
+                  _ -> fail $ "runZ3Horn: unexpected function " ++ show fn
+              | otherwise -> fail $ "runZ3Horn: unexpected function " ++ show fn)
```
RyanGlScott, line 323: It would be worth making these two errors have distinct messages so as to make it easier to tell them apart.

what4/src/What4/Expr/Builder.hs
```diff
@@ -946,6 +954,340 @@ evalBoundVars sym e vars exprs = do
                             }
   evalBoundVars' tbls sym e
 
+
+newtype ExprTransformer t (tp1 :: BaseType) (tp2 :: BaseType) a =
```
RyanGlScott, line 958: Most of the code added here in `What4.Expr.Builder` is undocumented, and as a result, I can't really follow what is going on here. Some additional comments would definitely be appreciated. (I'll trust your judgment to determine how much documentation is appropriate.)

what4/src/What4/Solver/Z3.hs
```diff
@@ -288,18 +302,54 @@ runZ3Horn sym log_data inv_fns horn_clauses = do
...
+
+        all_unsat <- and <$> mapM
+          (\clause -> do
+            defined_clause <- notPred sym =<< substituteSymFns sym defined_inv_fns clause
+            runZ3InOverride sym (log_data { logVerbosity = 2 }) [defined_clause] $ return . isUnsat)
+          horn_clauses
+
+        return $ if all_unsat then Sat defined_inv_fns else Unknown
+
+      _ -> return Unknown
```
RyanGlScott, line 334: It might be worth factoring this large subexpression out into its own function, as there is a lot of distance from the `  if do_bv_to_lia_transform then ...` part to the `... else return get_value_result` part.

what4/test/InvariantSynthesis.hs
```diff
@@ -148,6 +148,7 @@ main = do
       failureZ3 = "failure with older Z3 versions; upgrade to at least 4.8.9"
   defaultMain $ testGroup "Tests" $
     [ synthesis_test "int" intProblem "cvc5" CVC5.runCVC5SyGuS CVC5.runCVC5InOverride
-    , skipPre4_8_9 failureZ3 $ synthesis_test "int" intProblem "z3" Z3.runZ3Horn Z3.runZ3InOverride
+    , skipPre4_8_9 failureZ3 $ synthesis_test "int" intProblem "z3" (\sym -> Z3.runZ3Horn sym False) Z3.runZ3InOverride
     , synthesis_test "bv" bvProblem "cvc5" CVC5.runCVC5SyGuS CVC5.runCVC5InOverride
+    , skipPre4_8_9 failureZ3 $ synthesis_test "bv" bvProblem  "z3" (\sym -> Z3.runZ3Horn sym True) Z3.runZ3InOverride
```
RyanGlScott, line 153: I wonder if the name of this test should be something like `"bv->int"` instead of just `"bv"`. As I (very loosely) understand, passing `True` to `runZ3Horn` should convert a bitvector (`bv`) problem into a linear integer arithmetic problem (`int`) before solving it, yes?

RyanGlScott (commented):
I'm not clear on what the overall goal of this PR is, so I'd definitely appreciate some more documentation.

RyanGlScott:
The CI failure [here](https://github.com/GaloisInc/what4/actions/runs/7471159548/job/20330997873?pr=250) is most likely a result of using an old version of the GitHub Actions CI configuration. If you rebase on top of the `master` branch, I suspect that error will go away.

what4/src/What4/Expr/Builder.hs
```diff
@@ -2558,6 +2900,14 @@ instance IsExprBuilder (ExprBuilder t st fs) where
...
       bvUnary sym $ UnaryBV.uext u w
 
+    | Just (BaseIte _ _ c a b) <- asApp x
+    , Just a_bv <- asBV a
+    , Just b_bv <- asBV b = do
+      Just LeqProof <- return $ isPosNat w
+      a' <- bvLit sym w $ BV.zext w a_bv
+      b' <- bvLit sym w $ BV.zext w b_bv
+      bvIte sym c a' b'
+
```
kquick, line 2912: Am I correct in assuming all these general `BaseIte` are not related to LIA CHC and corresponding transformations and are just missing in our standard ExprBuilder logic?

kquick (commented):
Like Ryan, I'd like more general information and internal haddocks.  I am in general aligned with these changes, but would like a little more of that info.  We could probably use more tests as well, but I don't know the boundary between feasible v.s. practical for adding tests.

kquick (approved):

PR #252, opened 2024-01-16
More simplifications

andreistefanescu:
Remove the `arrayIte` simplification because it actually increases expression size, and combined with the muxing from the memory model can create very large expressions.


RyanGlScott (approved):
This was split out of https://github.com/GaloisInc/what4/pull/250, right?

I don't have a great intuition for which simplifications result in the most performant code, so I'm happy to defer to your judgment. If it's possible, it would be helpful to record some high-level comments somewhere about why `What4.Expr.Builder` deals with muxing the way it does.

andreistefanescu:
The simplifications are simply the result of me trying different things until the latest aws-lc proof works. There isn't any underlying reason to it. I agree, we should audit and document the existing simplification, and maybe have a more principled/exhaustive way of adding simplifications.

PR #256, opened 2024-03-01
Guard mux-pushing simplifications behind option

RyanGlScott:
Add `pushMuxOps` and `pushMuxOpsOption`. If this option is enabled, What4 will push certain `ExprBuilder` operations (e.g., `zext`) down to the branches of `ite` expressions. In some (but not all) circumstances, this can result in operations that are easier for SMT solvers to reason about.

### DIFF
--- a/what4/src/What4/Expr/Builder.hs
+++ b/what4/src/What4/Expr/Builder.hs
@@ -58,6 +58,7 @@ module What4.Expr.Builder
   , sbNonceExpr
   , curProgramLoc
   , unaryThreshold
+  , pushMuxOps
   , cacheStartSize
   , userState
   , exprCounter
@@ -76,6 +77,7 @@ module What4.Expr.Builder
     -- * configuration options
   , unaryThresholdOption
   , cacheStartSizeOption
+  , pushMuxOpsOption
   , cacheTerms
 
     -- * Expr
@@ -374,6 +376,12 @@ data ExprBuilder t (st :: Type -> Type) (fs :: Type)
           -- | The starting size when building a new cache
         , sbCacheStartSize :: !(CFG.OptionSetting BaseIntegerType)
 
+          -- | If enabled, push certain 'ExprBuilder' operations (e.g., @zext@)
+          -- down to the branches of @ite@ expressions. In some (but not all)
+          -- circumstances, this can result in operations that are easier for
+          -- SMT solvers to reason about.
+        , sbPushMuxOps :: !(CFG.OptionSetting BaseBoolType)
+
           -- | Counter to generate new unique identifiers for elements and functions.
         , sbExprCounter :: !(NonceGenerator IO t)
 
@@ -421,6 +429,9 @@ unaryThreshold = to sbUnaryThreshold
 cacheStartSize :: Getter (ExprBuilder t st fs) (CFG.OptionSetting BaseIntegerType)
 cacheStartSize = to sbCacheStartSize
 
+pushMuxOps :: Getter (ExprBuilder t st fs) (CFG.OptionSetting BaseBoolType)
+pushMuxOps = to sbPushMuxOps
+
 -- | Return a new expr builder where the configuration object has
 --   been "split" using the @splitConfig@ operation.
 --   The returned sym will share any preexisting options with the
@@ -456,9 +467,11 @@ exprBuilderFreshConfig sym =
      cfg <- CFG.initialConfig 0
               [ unaryThresholdDesc
               , cacheStartSizeDesc
+              , pushMuxOpsDesc
               ]
      unarySetting       <- CFG.getOptionSetting unaryThresholdOption cfg
      cacheStartSetting  <- CFG.getOptionSetting cacheStartSizeOption cfg
+     pushMuxOpsStartSetting <- CFG.getOptionSetting pushMuxOpsOption cfg
      CFG.extendConfig [cacheOptDesc gen storage_ref cacheStartSetting] cfg
      nonLinearOps <- newIORef 0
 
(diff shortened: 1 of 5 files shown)


what4/src/What4/Expr/Builder.hs
```diff
@@ -2913,17 +2948,20 @@ instance IsExprBuilder (ExprBuilder t st fs) where
...
-    , Just b_bv <- asBV b = do
-      Just LeqProof <- return $ isPosNat w
-      a' <- bvLit sym w $ BV.zext w a_bv
-      b' <- bvLit sym w $ BV.zext w b_bv
-      bvIte sym c a' b'
-
     | otherwise = do
-      Just LeqProof <- return $ testLeq (knownNat :: NatRepr 1) w
-      sbMakeExpr sym $ BVZext w x
+      pmo <- CFG.getOpt (sbPushMuxOps sym)
```
kquick, line 2952: Quick check: looks like you had to do this in 6-8 other locations as well.  Do you think it's reasonable to float this `getOpt` to the top before the primary case on expr so that the case pattern match branches can be largely as they were with just the `pmo` check?  I'm fine either way (based on the assumption that `getOpt` is inexpensive), but just wanted to call this out for you to make a decision on.

what4/src/What4/Expr/Builder.hs
```diff
@@ -650,6 +664,24 @@ unaryThresholdDesc = CFG.mkOpt unaryThresholdOption sty help (Just (ConcreteInte
...
         help = Just "Maximum number of values in unary bitvector encoding."
 
+------------------------------------------------------------------------
+-- Configuration option for controlling whether to push certain ExprBuilder
+-- operations (e.g., @zext@) down to the branches of @ite@ expressions.
+
+-- | If this option enabled, push certain 'ExprBuilder' operations (e.g.,
+-- @zext@) down to the branches of @ite@ expressions. In some (but not all)
+-- circumstances, this can result in operations that are easier for SMT solvers
+-- to reason about.
```
kquick, line 674: Suggest adding: "The expressions that may be pushed down are determined on a case-by-case basis in the `IsExprBuilder` instance code, but this control applies to all such push-down checks."

kquick (approved):
IIUC, this patch adds a config flag to control the push-down, and the push-down is originally added in a different patch that is also post 1.5.1.  Thus, the default for most clients (unless they had updated to a commit of what4 lying within these two changes) is unchanged and the new behavior only occurs when the new config is set.

My concern here is clients like Macaw that use pattern-based recognition (especially for end-of-block dispatching) that might be broken by this push-down behavior.  Unless you think this concern is unfounded, you might want to add this information to the PR description to help forensic efforts for any of this kind of downstream breakage folks might encounter.

Other than this concern and the help text, the other comments are just stylistic, so I'm approving and you are free to choose what changes you'd like to make or not make before merging.

what4/src/What4/Expr/Builder.hs
```diff
@@ -456,16 +467,19 @@ exprBuilderFreshConfig sym =
      cfg <- CFG.initialConfig 0
               [ unaryThresholdDesc
               , cacheStartSizeDesc
+              , pushMuxOpsDesc
               ]
      unarySetting       <- CFG.getOptionSetting unaryThresholdOption cfg
      cacheStartSetting  <- CFG.getOptionSetting cacheStartSizeOption cfg
+     pushMuxOpsStartSetting <- CFG.getOptionSetting pushMuxOpsOption cfg
```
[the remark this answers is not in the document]
RyanGlScott: Yes, good point—this is a copy-paste leftover from `cacheStartSetting` on the line above. I've renamed `pushMuxOpsStartSetting` to just `pushMuxOpsSetting`.

what4/src/What4/Expr/Builder.hs
```diff
@@ -650,6 +664,24 @@ unaryThresholdDesc = CFG.mkOpt unaryThresholdOption sty help (Just (ConcreteInte
...
+--
+-- This option is named \"backend.push_mux_ops\".
+pushMuxOpsOption :: CFG.ConfigOption BaseBoolType
+pushMuxOpsOption = CFG.configOption BaseBoolRepr "backend.push_mux_ops"
+
+-- | The 'CFG.ConfigDesc' for 'pushMuxOpsOption'.
+pushMuxOpsDesc :: CFG.ConfigDesc
+pushMuxOpsDesc = CFG.mkOpt pushMuxOpsOption sty help (Just (ConcreteBool False))
+  where sty = CFG.boolOptSty
+        help = Just "Maximum number of values in unary bitvector encoding."
```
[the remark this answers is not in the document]
RyanGlScott: Oops, this was left over from an incomplete copy-paste job. I've changed the description to something more appropriate.

what4/src/What4/Expr/Builder.hs
RyanGlScott: If I understand you correctly, you are suggesting to call `getOpt` at the beginning of each method's implementation to minimize the diff to the actual logic of the code itself? If so, I'm not sure that that would actually have the desired effect. In particular, each method of this `IsExprBuilder` instance is implemented using pattern guards, so in order to call `getOpt` before the pattern guards, we'd likely need to do something like this:

```hs
  bvZext sym w x = do
    pmo <- CFG.getOpt (sbPushMuxOps sym)
    if | Just xv <- asBV x = do ...
       | Just (BVZext _ y) <- asApp x = do ...
       |  ...
```

But using `MultiWayIf` (or a similar refactoring) here would cause the diff to be just as large (if not larger) than it is currently. As such, I opted to localize the `getOpt` calls to the parts of the code that actually make use of the option.

RyanGlScott (commented):
> My concern here is clients like Macaw that use pattern-based recognition (especially for end-of-block dispatching) that might be broken by this push-down behavior. Unless you think this concern is unfounded, you might want to add this information to the PR description to help forensic efforts for any of this kind of downstream breakage folks might encounter.

To be clear, Macaw isn't yet making use of the new behavior, and by the time Macaw bumps its What4 submodule, the new behavior will be guarded behind an opt-in flag. As a result, Macaw's behavior won't change unless it specifically opts into the new behavior. (Indeed, this was the whole point of guarding this behavior behind a flag, as the new behavior proved problematic for certain use cases, such as in certain SAW proofs.)

PR #258, opened 2024-04-03
Add `bv{Zero,One}` helpers, hlint config

### LINKED ISSUE #257: Helpers for constructing bitvectors 0 and 1
It's common to construct literal bitvectors representing 0 or 1:

- https://github.com/search?q=repo%3AGaloisInc%2Fwhat4%20%2FbvLit.%2Bzero%2F&type=code
- https://github.com/search?q=repo%3AGaloisInc%2Fcrucible%20%2FbvLit.%2Bzero%2F&type=code
- https://github.com/search?q=repo%3AGaloisInc%2Fwhat4+%2FbvLit.%2Bone%2F&type=code
- https://github.com/search?q=repo%3AGaloisInc%2Fcrucible+%2FbvLit.%2Bone%2F&type=code

The best way to do that in the current interface is:
```haskell
bvLit sym w (BV.zero w)
bvLit sym w (BV.one w)
```
This is a bit of a pain, since you have to repeat the width repr, and to use functions from two separate packages. Instead, it might be nice to export
```
bvZero, bvOne :: IsSymInterface sym => NatRepr w -> IO (SymBV sym w)
```
The names also make the intent more obvious.

langston-barrett:
Fixes #257. The HLint configuration only checks that these helpers are used where appropriate. I used it to find places where they would be useful. It may also serve as a template for downstream repos. I added HLint checking to CI as well.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -12,9 +12,3 @@
     name: Use bvOne
     lhs: "What4.Interface.bvLit sym w (Data.BitVector.Sized.one w)"
     rhs: 'What4.Interface.bvOne sym w'
-
-# TODO:
-# - error:
-#     name: Use PP.viaShow
-#     lhs: "Prettyprinter.pretty (show x)"
-#     rhs: 'Prettyprinter.viaShow x'
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -1,75 +1,20 @@
-# HLint configuration file
-# https://github.com/ndmitchell/hlint
-##########################
+# HLint's default suggestions are opinionated, so we disable all of them by
+# default and enable just a small subset we can agree on.
 
-- modules:
-  - {name: [Data.Set, Data.HashSet], as: Set} # if you import Data.Set qualified, it must be as 'Set'
-  - {name: [Data.List], as: List}
-  - {name: [Data.Sequence], as: Seq}
+- ignore: {}  # ignore all
 
-# Add custom hints for this project
-#
-# Will suggest replacing "wibbleMany [myvar]" with "wibbleOne myvar"
-# - error: {lhs: "wibbleMany [x]", rhs: wibbleOne x}
+- error:
+    name: Use bvZero
+    lhs: "What4.Interface.bvLit sym w (Data.BitVector.Sized.zero w)"
+    rhs: 'What4.Interface.bvZero sym w'
 
-# We should use "panic", not "error".
+- error:
+    name: Use bvOne
+    lhs: "What4.Interface.bvLit sym w (Data.BitVector.Sized.one w)"
+    rhs: 'What4.Interface.bvOne sym w'
+
+# TODO:
 # - error:
-#     lhs: "error x"
-#     rhs: 'panic "nameOfFunction" [x, "more lines of details"]'
-
-# TODO: specialize these to the modules they are needed
-- ignore: {name: 'Use :'}
-- ignore: {name: Avoid lambda using `infix`}
-- ignore: {name: Avoid lambda}
-- ignore: {name: Avoid restricted qualification}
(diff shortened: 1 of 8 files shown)


what4/src/What4/Interface.hs
```diff
@@ -3294,3 +3298,16 @@ data Statistics
...
+
+-- | An alias for 'minUnsignedBv'.
+--
+-- Useful in contexts where you want to convey the zero-ness of the value more
+-- than its minimality.
+bvZero :: (1 <= w, IsExprBuilder sym) => sym -> NatRepr w -> IO (SymBV sym w)
+bvZero = minUnsignedBV
+
+bvOne :: (1 <= w, IsExprBuilder sym) => sym -> NatRepr w -> IO (SymBV sym w)
+bvOne sym w = bvLit sym w (BV.one w)
```
RyanGlScott, line 3313: Haddocks?

what4/src/What4/Interface.hs
```diff
@@ -180,6 +180,10 @@ module What4.Interface
     -- * Exceptions
   , InvalidRange(..)
 
+    -- * Bitvector utilities
+  , bvZero
+  , bvOne
```
RyanGlScott, line 185: Perhaps it makes more sense to include this under the "Utility combinators" Haddock section above as a sub-section.

.hlint.yaml
```diff
@@ -1,75 +1,20 @@
...
-- ignore: {name: Use uncurry}
-- ignore: {name: Use unless}
-- ignore: {name: Use unwords}
-- ignore: {name: Use void}
-- ignore: {name: Use when}
+# TODO:
+# - error:
+#     name: Use PP.viaShow
+#     lhs: "Prettyprinter.pretty (show x)"
+#     rhs: 'Prettyprinter.viaShow x'
```
RyanGlScott, line 20: Delete this?

what4/src/What4/Interface.hs
```diff
@@ -1117,7 +1121,7 @@ class ( IsExpr (SymExpr sym), HashableF (SymExpr sym), HashableF (BoundVar sym)
   -- | Return the bitvector of the desired width with all 0 bits;
   --   this is the minimum unsigned integer.
   minUnsignedBV :: (1 <= w) => sym -> NatRepr w -> IO (SymBV sym w)
-  minUnsignedBV sym w = bvLit sym w (BV.zero w)
+  minUnsignedBV sym w = bvZero sym w
```
RyanGlScott, line 1124: Eek, I don't think you meant to change this part! `minUnsigned` and `bvZero` are now defined mutually in terms of each other, so calling either function will result in an infinite loop. (The CI can testify to this.)

I suppose this means that the `.hlint.yaml` file will need to carve out an exception for this module.

langston-barrett: Whoops, yep! This was a regex search/replace :sweat:

langston-barrett: For some reason, `hlint` doesn't seem to tag this nor `bvOne`. That seems fine, false negatives aren't a big problem for this kind of lint.

.hlint.yaml
langston-barrett: Yeah, I was thinking we might want to do this in a follow-up, but can remove it for now.

what4/src/What4/Interface.hs
langston-barrett: Good call

what4/src/What4/Interface.hs
langston-barrett: For once, I had already written them, I just forgot to push that commit :grin: I'm learning!

RyanGlScott (approved):

Issue #259, opened 2024-04-17
Concretization


langston-barrett:
The [`What4.Utils.ResolveBounds.BV.resolveSymBv`](https://hackage.haskell.org/package/what4-1.5.1/docs/What4-Utils-ResolveBounds-BV.html#v:resolveSymBV) function can be applied to a symbolic bitvector to (1) check if it is actually only equal to a single value or (2) get upper and lower bounds for it otherwise. There are reasonable use-cases that only require (1), which also has a much more reasonable upper bound on the number of SMT queries it issues. 

Furthermore, we can probably easily generalize (1) to other base types. Here's the code:

https://github.com/GaloisInc/what4/blob/30309b5d86179c2d4182f68ccf36e8dd5f68360c/what4/src/What4/Utils/ResolveBounds/BV.hs#L109-L136

To generalize, we can replace:

- `asBV` -> `asConcrete`
- `bvLit` -> `concreteToSym`
- `bvEq` -> `isEq`

RyanGlScott:
See also the `Concretize` machinery in [`Data.Macaw.Symbolic.Concretize`](https://github.com/GaloisInc/macaw/blob/442b0e9f00c1aa9dbe71a62edcb4e0039452691b/symbolic/src/Data/Macaw/Symbolic/Concretize.hs#L107-L120), which attempts to provide a more generic concretization interface akin to what you describe above as (1). (Despite living in `macaw-symbolic`, none of the code in that module is Macaw-specific.)

langston-barrett:
See also [`Pate.Verification.Concretize`](https://github.com/GaloisInc/pate/blob/c43542818be7780264d8a2552bfeba1cffba2902/src/Pate/Verification/Concretize.hs)

langston-barrett:
See also [`crux-mir`'s take](https://github.com/GaloisInc/crucible/blob/45648cc1ebc834fea38f126e6cac869101189fd0/crux-mir/src/Mir/Overrides.hs#L128-L182)

PR #260, opened 2024-04-17
Remove `NatRepr` arg from `resolveSymBV`

langston-barrett:
It's not necessary; it can be recovered from the bitvector itself.


RyanGlScott (commented):
I agree with this in principle, although I wonder if this should be done in tandem with other (potentially API-breaking) changes discussed in #259.

langston-barrett:
Sounds good to me, I'll reopen after implementing the features discussed there!

PR #264, opened 2024-05-24
Concretization and unique concretization

### LINKED ISSUE #259: Concretization

langston-barrett:
Fixes #259.

### DIFF
--- a/what4/src/What4/Concretize.hs
+++ b/what4/src/What4/Concretize.hs
@@ -82,4 +82,4 @@ concretize sym sp val =
               case msat' of
                 WSat.Unknown -> pure $ Left SolverUnknown -- Total failure
                 WSat.Sat _mdl -> pure $ Left MultipleModels  -- There are multiple models
-                WSat.Unsat {} -> pure (Right concVal) -- There is a single concrete result
+                WSat.Unsat {} -> pure $ Right concVal -- There is a single concrete result
--- a/what4/src/What4/Concretize.hs
+++ b/what4/src/What4/Concretize.hs
@@ -69,7 +69,7 @@ data UniqueConcretizationFailure
   = GroundingFailure ConcretizationFailure
   | MultipleModels
     -- ^ There are multiple possible models for the expression, which means it
-    -- is truly symbolic and therefore unable to be concretized.
+    -- is truly symbolic and therefore unable to be uniquely concretized.
   deriving Show
 
 -- | Attempt to resolve the given 'WI.SymExpr' to a unique concrete value using
--- a/what4/src/What4/Concretize.hs
+++ b/what4/src/What4/Concretize.hs
@@ -6,6 +6,11 @@
 -- License          : BSD3
 -- Maintainer       : Langston Barrett <langston@galois.com>
 -- Stability        : provisional
+--
+-- In our terminology, concretization is the process of (1) obtaining a
+-- model from an SMT solver and (2) requesting the value of a particular set
+-- of symbolic expressions in said model. The operation (2) alone is called
+-- "grounding", see "What4.GroundEval".
 -----------------------------------------------------------------------
 
 {-# LANGUAGE DataKinds #-}
--- a/what4/src/What4/Concretize.hs
+++ b/what4/src/What4/Concretize.hs
@@ -35,8 +35,12 @@ data GroundingFailure
     -- due to the initial assumptions in scope being unsatisfiable.
   deriving Show
 
--- | Get a 'WEG.GroundValue' for a 'WI.SymExpr' by asking the solver for a
--- model.
+-- | Get a 'WEG.GroundValue' for a 'WI.SymExpr' by asking an online solver for
+-- a model.
+--
+-- In contrast with 'concretize', this function returns the value of the
+-- 'WI.SymExpr' in just one of potentially many distinct models. See the Haddock
+-- on 'concretize' for a further comparison.
 groundFromModel ::
   ( sym ~ WEB.ExprBuilder scope st fs
   , WPO.OnlineSolver solver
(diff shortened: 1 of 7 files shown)


what4/src/What4/Concretize.hs
```diff
@@ -0,0 +1,95 @@
...
+  -- | The symbolic backend
+  sym ->
+  WPO.SolverProcess scope solver ->
+  -- | The symbolic term to concretize
+  WI.SymExpr sym tp ->
+  IO (Either ConcretizationFailure (WEG.GroundValue tp))
+concretize sym sp val =
+  case WI.asConcrete val of
+    Just val' ->
+      case WEG.fromConcrete val' of
```
langston-barrett, line 64: Would probably be better to avoid this potential failure by implementing `asGround :: SymExpr -> GroundValue` instead.

what4/src/What4/Expr/GroundEval.hs
```diff
@@ -84,6 +85,20 @@ type family GroundValue (tp :: BaseType) where
...
+  case exprType x of
+    BaseBoolRepr       -> asConstantPred x
+    BaseIntegerRepr    -> asInteger x
+    BaseRealRepr       -> asRational x
+    BaseStringRepr _si -> asString x
+    BaseComplexRepr    -> asComplex x
+    BaseBVRepr _w       -> asBV x
+    BaseFloatRepr _fpp  -> asFloat x
+    BaseStructRepr _   -> asStruct x >>= traverseFC (fmap GVW . asGround)
+    BaseArrayRepr _idx _tp -> Nothing
```
RyanGlScott, line 100: This function is strikingly similar to [`What4.Interface.asConcrete`](https://github.com/GaloisInc/what4/blob/c4573936f98d1deb0423af0b489ca63b34aaf4c8/what4/src/What4/Interface.hs#L3202-L3218). Is it feasible to define `asConcrete` in terms of `asGround`? If not, it would be worth mentioning `asConcrete` in `asGround`'s Haddocks, similarly to how we do it fro `groundToSym` (which mentions `What4.Interface.concreteToSym`).

what4/src/What4/Concretize.hs
```diff
@@ -0,0 +1,85 @@
...
+concretize ::
+  ( sym ~ WEB.ExprBuilder scope st fs
+  , WPO.OnlineSolver solver
+  ) =>
+  -- | The symbolic backend
+  sym ->
+  WPO.SolverProcess scope solver ->
+  -- | The symbolic term to concretize
+  WI.SymExpr sym tp ->
+  IO (Either ConcretizationFailure (WEG.GroundValue tp))
```
RyanGlScott, line 56: The motivation for #259 (as written in the issue description) is to use this function to implement [`resolveSymBV`](https://hackage.haskell.org/package/what4-1.5.1/docs/What4-Utils-ResolveBounds-BV.html#v:resolveSymBV). Do you still plan to do this?

what4/src/What4/Concretize.hs
```diff
@@ -0,0 +1,85 @@
...
+            WPO.inNewFrame sp $ do
+              injectedConcVal <- WEG.groundToSym sym (WI.exprType val) concVal
+              eq <- WI.isEq sym val injectedConcVal
+              block <- WI.notPred sym eq
+              WPS.assume (WPO.solverConn sp) block
+              msat' <- WPO.checkAndGetModel sp "Concretize value (with blocking clause)"
+              case msat' of
+                WSat.Unknown -> pure $ Left SolverUnknown -- Total failure
+                WSat.Sat _mdl -> pure $ Left MultipleModels  -- There are multiple models
+                WSat.Unsat {} -> pure (Right concVal) -- There is a single concrete result
```
RyanGlScott, line 85: Style consistency:

```suggestion
                WSat.Unknown -> pure $ Left SolverUnknown -- Total failure
                WSat.Sat _mdl -> pure $ Left MultipleModels  -- There are multiple models
                WSat.Unsat {} -> pure $ Right concVal -- There is a single concrete result
```

what4/src/What4/Expr/GroundEval.hs
RyanGlScott, line 100: I think we can do a little better here than always returning `Nothing` for arrays. See the corresponding case in `asConcrete`:

https://github.com/GaloisInc/what4/blob/c4573936f98d1deb0423af0b489ca63b34aaf4c8/what4/src/What4/Interface.hs#L3213-L3218

This can return `Just` if the all the array elements are a constant value, which is a slight improvement. (We'd need to store the result in [`ArrayConcrete`](https://hackage-content.haskell.org/package/what4-1.7/docs/What4-Expr-GroundEval.html#v:ArrayConcrete) instead of `ConcreteArray`.)

what4/src/What4/Concretize.hs
```diff
@@ -0,0 +1,108 @@
...
+data GroundingFailure
+  = SolverUnknown
+    -- ^ Querying the SMT solver yielded @UNKNOWN@.
+  | UnsatInitialAssumptions
+    -- ^ Querying the SMT solver for an initial model of the expression failed
+    -- due to the initial assumptions in scope being unsatisfiable.
+  deriving Show
+
+-- | Get a 'WEG.GroundValue' for a 'WI.SymExpr' by asking the solver for a
+-- model.
```
RyanGlScott, line 39: Since `groundFromModel` and `concretize` have nearly the same type signature, it would be worth contrasting each function to make it clearer when you'd pick one over the other.

RyanGlScott (approved):
LGTM, modulo a few minor comments.

what4/src/What4/Expr/GroundEval.hs
RyanGlScott: I think the TODO is more referring to the fact that this implementation is rather conservative and will miss concrete arrays such as `[0 :-> True; _ :-> False]`, i.e., arrays where most indices map to a constant, but with a finite number of indices mapping to a different value. I suppose it wouldn't hurt to clarify the TODO to reflect this, although I am not sure if we can necessarily do better given the current `what4` interface.

what4/src/What4/Concretize.hs
langston-barrett: Yep, good point. Done!

langston-barrett:
I realized that calling this "concretization" conflicts with how that term is used in [`Lang.Crucible.Concretize`](https://github.com/GaloisInc/crucible/blob/2e0e02b95d7d057eae13945a9e0875994c55757b/crucible/src/Lang/Crucible/Concretize.hs). I'm seeking input on the naming scheme here before continuing.

Issue #265, opened 2024-05-31
CVC5 fails to parse `what4`-generated SMT-LIB for constructing and selecting tuples


RyanGlScott:
This program:

```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE TypeApplications #-}
module Main where

import qualified Data.Parameterized.Context as Ctx
import qualified Data.Parameterized.Nonce as N
import qualified Data.Parameterized.Some as Some
import qualified System.IO as IO
import qualified What4.Config as WC
import qualified What4.Expr as WE
import qualified What4.Expr.Builder as WEB
import qualified What4.Expr.GroundEval as WEG
import qualified What4.Interface as WI
import qualified What4.Solver as WS

opts :: [WC.ConfigDesc]
opts = WS.cvc5Options

adapter :: WS.SolverAdapter st
adapter = WS.cvc5Adapter

main :: IO ()
main = do
  Some.Some ng <- N.newIONonceGenerator
  sym <- WEB.newExprBuilder WEB.FloatIEEERepr WE.EmptyExprBuilderState ng
  WC.extendConfig opts (WI.getConfiguration sym)

  let w = WI.knownNat @32
  ufn <- WI.freshTotalUninterpFn sym (WI.safeSymbol "ufn")
           (Ctx.Empty Ctx.:> WI.BaseStructRepr (Ctx.Empty Ctx.:> WI.BaseBVRepr w))
           (WI.BaseStructRepr (Ctx.Empty Ctx.:> WI.BaseBVRepr w))
  fldArg <- WI.freshConstant sym (WI.safeSymbol "fldArg") (WI.BaseBVRepr w)
  structArg <- WI.mkStruct sym (Ctx.Empty Ctx.:> fldArg)
  structRes <- WI.applySymFn sym ufn (Ctx.Empty Ctx.:> structArg)
  fldRes <- WI.structField sym structRes Ctx.baseIndex
  fldResNeg <- WI.bvNeg sym fldRes
  p <- WI.bvEq sym fldRes fldResNeg

  WS.solver_adapter_check_sat adapter sym
    WS.defaultLogData{WS.logHandle = Just IO.stdout} [p] $ \_result ->
    case _result of
      WS.Sat (ge, _) -> do
        putStrLn "Satisfiable, with model:"
        v <- WEG.groundEval ge fldArg
        putStrLn $ "fldArg := " ++ show v
[...]

PR #266, opened 2024-05-31
CVC5: Don't use Tuple workaround when declaring structs

RyanGlScott:
Due to a bug in CVC4 (https://github.com/cvc5/cvc5/issues/3402), the What4 bindings to CVC4 avoid user-defined datatypes and instead use CVC4's built-in `Tuple` datatype. This same `Tuple` workaround was applied to the CVC5 bindings (added in https://github.com/GaloisInc/what4/pull/204), but they don't really work properly, as CVC4 and CVC5 use completely different syntaxes for tuples. See https://github.com/GaloisInc/what4/issues/265.

We could attempt to update the CVC5 bindings to use the modern tuple syntax, but this is somewhat non-trivial, as the syntax for 1-tuples changed in `cvc5-1.0.9` from `Tuple` to `UnitTuple` (see https://github.com/cvc5/cvc5/pull/10012). As such, we would have to generate different code for different versions of CVC5, which would be unpleasant.

Thankfully, we need not worry about tuples at all when it comes to CVC5. This is because the original CVC4 issue (https://github.com/cvc5/cvc5/issues/3402) does not affect CVC5. As such, we can drop the tuple workaround entirely and just generate user-defined datatypes to represent What4 structs, exactly like the other solver bindings do. This means that we can fix https://github.com/GaloisInc/what4/issues/265 by simply deleting code, which is nice.

In addition, this PR tweaks the `expr-builder-smtlib2` test suite to _actually_ run its tests with both CVC4 and CVC5. The `0-tuple` and `1-tuple` tests catch the exact bug described in #265, so these serve as a regression test for #265 when run with CVC5.

Fixes https://github.com/GaloisInc/what4/issues/265.

### DIFF
--- a/.github/workflows/gen_matrix.pl
+++ b/.github/workflows/gen_matrix.pl
@@ -78,7 +78,7 @@ main_version(z3, "4_8_14").
 main_version(yices, "2_6_4").
 main_version(stp, "2_3_3").
 main_version(cvc4, "1_8").
-main_version(cvc5, "1_0_2").
+main_version(cvc5, "1_1_2").
 main_version(bitwuzla, "0_3_0").
 main_version(boolector, "3_2_2").
 main_version(abc, "2021_12_30").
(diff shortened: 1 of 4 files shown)


RyanGlScott:
I don't think so. CVC5 does in fact use structs—that fact remains unchanged by this PR. The only difference is how it implements structs under the hood: it used tuples before, but now it uses user-defined datatypes.

PR #277, opened 2025-02-03
Tests and documentation for normalization of boolean values

langston-barrett:
- Introduce a `ConjMap` newtype around `BoolMap` for readability of pattern matches and type signatures
- Document the annotation on `BoolMap`
- Document and introduce tests for normal forms of boolean expressions


langston-barrett:
I don't understand these CI failures:
```
    Z3
      Test itself is valid and completes (2.27 s): OK (292.02s)
      Test runs past timeout:                      OK (1.40s)
      Test with goal timeout (0.25 s):             FAIL (expected: goal timeouts feature not effective) (1.40s)
        test/OnlineSolverTest.hs:485:
        solver goal timeout didn't occur (expected failure)
    CVC4
      Test itself is valid and completes (7.5 s):  OK (8.13s)
      Test runs past timeout:                      FAIL (2.80s)
        test/OnlineSolverTest.hs:463:
        Solver check query not interruptible (2.8026422220000002 s > expected 1.4 s)
        Use -p '/CVC4.Test runs past timeout/' to rerun this test only.
```

langston-barrett:
The CI failure is #278.

what4/src/What4/Expr/BoolMap.hs
```diff
@@ -179,3 +208,60 @@ reversePolarities (BoolMap m) = BoolMap $! fmap negatePolarity m
...
+pattern ConjTrue :: ConjMapView f
+pattern ConjTrue = ConjMapView BoolMapUnit
+
+pattern ConjFalse :: ConjMapView f
+pattern ConjFalse = ConjMapView BoolMapDualUnit
+
+pattern Conjuncts :: NonEmpty (f BaseBoolType, Polarity) -> ConjMapView f
+pattern Conjuncts ts = ConjMapView (BoolMapTerms ts)
+
+{-# COMPLETE ConjTrue, ConjFalse, Conjuncts #-}
```
RyanGlScott, line 236: I have to admit that I find it strange to introduce a set of pattern synonyms that are (as far as I can tell) completely isomorphic to the original data constructors for `BoolMapView`. I wonder whether it would be less confusing to simply rename `BoolMapView` and its data constructors to `ConjMapView` _et al._ (That would be a breaking API change, yes, but this PR already changes the API by changing the type of `ConjPred`'s field...)

Is there value in keeping both `BoolMapView` and `ConjMapView` around?

what4/src/What4/Expr/BoolMap.hs
```diff
@@ -179,3 +208,60 @@ reversePolarities (BoolMap m) = BoolMap $! fmap negatePolarity m
 removeVar :: OrdF f => BoolMap f -> f BaseBoolType -> BoolMap f
 removeVar InconsistentMap _ = InconsistentMap
 removeVar (BoolMap m) x = BoolMap (AM.delete (Wrap x) m)
+
+--------------------------------------------------------------------------------
+-- ConjMap
+
+-- No idea why `coerce` needs these the explicit type applications in this
+-- section...
```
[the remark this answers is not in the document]
langston-barrett: Ah, thanks! I'll add a link.

what4/src/What4/Expr/BoolMap.hs
langston-barrett: Ah yes, great question. I assumed that we would want to keep `BoolMap{,View}` around, because it is equally capable of representing disjunctions. We don't happen to use it that way at the moment, but it doesn't seem strange to think we might in the future. In a context where we keep them, then the patterns are just for readability at the sites of pattern matches. For instance, looking at [this part of the diff](https://github.com/GaloisInc/what4/pull/277/files#diff-42406ab44e1a62a7aa863cfe2f2f5ed35cd21cdd6bd63569ae95b8e1db9a84faL2344), it previously wasn't immediately clear that the strings were correct - you had to know what `BoolMapUnit` meant, both overall, and whether this particular `BoolMap` represented a conjunction or a disjunction. Now, it is obviously correct.

I probably lean towards keeping `BoolMap`, but I'm fine with renaming it and adding a note that the implementation doesn't actually assume it is a conjunction instead if you think it would be better.

A third option would be to keep `BoolMap`, but to just specialize `{b,B}oolMapView` into `{c,C}onjMapView`. Perhaps we could introduce a `{d,D}isjMapView` to retain that functionality as well. I'm also fine with this option.

>  (That would be a breaking API change, yes, but this PR already changes the API by changing the type of ConjPred's field...)

Yeah, it seems like potentially a mistake to be exporting `Expr`'s constructors? It's a huge, complex datatype and we might want to experiment with different representations/normal forms. For instance, it was a breaking change to introduce `BoolMap` and `WeightedSum`, but both are presumably big improvements over what came before.

RyanGlScott: Ah, I hadn't considered that `BoolMap` could just as well represent a disjunction. Indeed, the documentation for `BoolMap` makes this more obvious... I should have read that first. (Oops.)

In that case, I think I am fine with having `ConjMapView` be separate. This might also be a good opportunity to improve the documentation for `BoolMapView` to make it more obvious that it could represent a conjunction or a disjunction.

> Yeah, it seems like potentially a mistake to be exporting `Expr`'s constructors?

Perhaps so. At the same time, there are downstream libraries that are writing their own optimization passes over `Expr`s (e.g., [this code](https://github.com/GaloisInc/pate/blob/7e7f0551775fcd7e19eeb8270e863cc6a86cc9b8/src/What4/ExprHelpers.hs#L1327-L1547) in `pate`). It is what it is.

langston-barrett: > In that case, I think I am fine with having ConjMapView be separate. This might also be a good opportunity to improve the documentation for BoolMapView to make it more obvious that it could represent a conjunction or a disjunction.

:+1:

> Perhaps so. At the same time, there are downstream libraries that are writing their own optimization passes over Exprs (e.g., [this code](https://github.com/GaloisInc/pate/blob/7e7f0551775fcd7e19eeb8270e863cc6a86cc9b8/src/What4/ExprHelpers.hs#L1327-L1547) in pate). It is what it is.

Yeah, to me that seems suboptimal - Vastly simplifying for a moment, functionality is either specific to a downstream project (and in that case would probably live in that project's module hierarchy) or it is generic and should live in the upstream project. That being said, there are of course cases where one needs to make an exception, and those should probably be covered by having `*.Internal` modules where virtually everything is exported with no promises of stability. If we want to maintain stability promises on What4's API and not bump major versions with regularity, What4 2.0 should probably export a whole heck of a lot less.

RyanGlScott: Aye, having clearer separation between stable and internal (i.e., unstable) parts of the API would be ideal.

what4/CHANGES.md
```diff
@@ -1,3 +1,7 @@
+# next
+
+* The `BoolMap` parameter of `ConjPred` is now a `ConjMap`.
```
RyanGlScott, line 3: `ConjMap` is a new addition to this version of `what4`, so it would be worth adding a brief explanation of what `ConjMap` is as well.

what4/src/What4/Expr/BoolMap.hs
```diff
@@ -179,3 +207,61 @@ reversePolarities (BoolMap m) = BoolMap $! fmap negatePolarity m
 removeVar :: OrdF f => BoolMap f -> f BaseBoolType -> BoolMap f
 removeVar InconsistentMap _ = InconsistentMap
 removeVar (BoolMap m) x = BoolMap (AM.delete (Wrap x) m)
+
+--------------------------------------------------------------------------------
+-- ConjMap
+
+-- | A 'BoolMap' representing a conjunction.
```
RyanGlScott, line 214: If you could, please also add some documentation to `BoolMapView` stating that it is equally capable of representing conjunctions and disjunctions, and that if you know you are dealing with a conjunction, you may want to consider using `ConjMap` and friends for clarity.

what4/src/What4/Expr/BoolMap.hs
```diff
@@ -179,3 +207,61 @@ reversePolarities (BoolMap m) = BoolMap $! fmap negatePolarity m
...
+
+-- | Deconstruct the given 'ConjMap' for later processing
+viewConjMap :: forall f. ConjMap f -> ConjMapView f
+viewConjMap =
+  -- The explicit type annotations on `coerce` are likely necessary because of
+  -- https://gitlab.haskell.org/ghc/ghc/-/issues/21003
+  coerce @(BoolMap f -> BoolMapView f) @(ConjMap f -> ConjMapView f) viewBoolMap
+{-# INLINE viewConjMap #-}
+
+addConjunct ::
```
RyanGlScott, line 242: Haddocks?

what4/src/What4/Expr/BoolMap.hs
```diff
@@ -179,3 +207,61 @@ reversePolarities (BoolMap m) = BoolMap $! fmap negatePolarity m
...
+addConjunct =
+  -- The explicit type annotations on `coerce` are likely necessary because of
+  -- https://gitlab.haskell.org/ghc/ghc/-/issues/21003
+  coerce
+    @(f BaseBoolType -> Polarity -> BoolMap f -> BoolMap f)
+    @(f BaseBoolType -> Polarity -> ConjMap f -> ConjMap f)
+    addVar
+{-# INLINE addConjunct #-}
+
+evalConj :: Applicative m => (f BaseBoolType -> m Bool) -> ConjMap f -> m Bool
```
RyanGlScott, line 258: Haddocks?

what4/src/What4/Expr/Builder.hs
```diff
@@ -19,6 +19,20 @@ threads may still clobber state others have set (e.g., the current
...
+and concrete as possible. In particular, here are a few considerations for
+boolean-typed expressions:
+
+* Disjunctions are implicitly represented as negated conjunctions
+* Conjunctions are represented via 'BM.ConjMap' (see docs on that type)
+* @xor@ is represented as the negation of equality
+
+Boolean expressions are expected to be somewhat normalized at all times.
+For example, there should never be a double negation (nested 'NotPred').
+See @isNormal@ in @test/Bool.hs@ for the exact expectations.
```
RyanGlScott, line 35: I would argue that we should spell out the expectations for how boolean normalization should work here in the library Haddocks, not tucked away in the comments of a test suite. Do you agree?

what4/test/Bool.hs
```diff
@@ -0,0 +1,280 @@
...
+  Either String ()
+isNormalConjunct sym expr pol =
+  case expr of
+    BoolExpr {} -> Left "boolean literal inside conjunction"
+    BoundVarExpr {} -> Right ()
+    AppExpr ae ->
+      case appExprApp ae of
+        NotPred {} -> Left "not should be expressed via polarity"
+        -- This must be an OR, if it is an AND it should be combined with
+        -- its parent
```
RyanGlScott, line 212: If we move the explanation for how boolean normalization works to the library Haddocks, then we could leave comments next to individual lines of code here in `test/Bool.hs` to justify why certain checks are performed. For instance, it's not obvious to me why this case should constitute an error, and I would have to trawl through the implementation in the library to figure out why this is a reasonable expectation. (Similarly for other checks in `test/Bool.hs`.)

what4/test/Bool.hs
```diff
@@ -0,0 +1,280 @@
...
+import Hedgehog.Internal.Gen qualified as HG
+import Hedgehog.Internal.Property qualified as HG
+import Test.Tasty.Hedgehog qualified as THG
+import Test.Tasty qualified as T
+import What4.Expr.BoolMap qualified as BM
+import What4.Expr.Builder
+import What4.Expr (EmptyExprBuilderState(EmptyExprBuilderState))
+import What4.Interface
+
+data BExpr a
```
RyanGlScott, line 31: Worth a comment here saying what the purpose of this `BExpr` data type is. My understanding is that it is meant to provide a simple reference implementation of how boolean evaluation should work that we can compare against the more complication evaluation scheme in `what4` itself. Is this correct?

what4/src/What4/Expr/Builder.hs
langston-barrett: If you agree with the direction of #282, I might like to defer this to an issue for now. I think that this is already an improvement over the status quo (no documentation, no tests), and the solution to #282 will change the invariants.

what4/test/Bool.hs
langston-barrett: > to justify why certain checks are performed

Broadly, the justification is the same for all of them: to reduce the sizes of terms, to keep terms more concrete, and (in the service of the first two) to keep terms in normal forms (e.g., so that they can be compared for equality, and rewrite rules can assume normal forms as inputs). However, we could certainly note *how* each invariant contributes to these goals.

what4/test/Bool.hs
langston-barrett: Good call, done!

what4/src/What4/Expr/Builder.hs
RyanGlScott: I'm OK with deferring to an issue, as well as the general direction of #282.

langston-barrett: :+1: #284

what4/test/Bool.hs
langston-barrett: Let's do this as part of #284.

langston-barrett:
Rebased onto #283 for CI fixes.

RyanGlScott (approved):
LGTM, although the CI is failing now for an unrelated reason: the code uses `ImportQualifiedPost`, which is only available in GHC 8.10 or later. We could drop support for GHC 8.8 to support this, although this might be better done in a separate patch.

langston-barrett:
Removed, thanks!

Issue #278, opened 2025-02-04
CVC 1.8 timeout tests failing in CI


langston-barrett:
https://github.com/GaloisInc/what4/actions/runs/13142662183

RyanGlScott:
It is strange that this failure only surfaced recently. What's more, I am able to reproduce the errors locally on my machine.

A wild shot in the dark: perhaps the performance characteristics of the CI runners were once such that these CVC4-related tests would take long enough that the timeout could reliably kick in, but the performance characteristics of the CI runners are now such that this no longer occurs. (I don't have any evidence for this claim, but I'm not sure what else might be causing this.) Either way, we should redesign these tests so that they aren't quite so sensitive to the performance characteristics of individual machines.

PR #283, opened 2025-02-05
Disable timeout test for CVC4 1.8

langston-barrett:
Work around #278 for now.


langston-barrett:
Hmm, one of the variants is failing with
```
    int mux:                                 FAIL (2.02s)
        ✗ int mux failed 
          after 500 tests.
```
I wonder if this is just a failure case that is rarely hit by the random generator?

RyanGlScott (approved):
When it rains, it pours. In nothing else, I'm happy to see at least _some_ of the flaky tests become hidden :)

> I wonder if this is just a failure case that is rarely hit by the random generator?

I think so, since restarting the job makes that test case pass.

PR #288, opened 2025-02-26
SMTLib2: only send exit command if solver process is active

danmatichuk:
this avoids issues where the solver process was
killed via `What4.Protocol.Online.killSolver`
(e.g. due to a forced timeout)


kquick (approved):
Good find, thanks!

RyanGlScott (commented):
Can you mention this bug fix in the changelog?

PR #293, opened 2025-03-20
Prepare release 1.7.0

sauclovian-g:


sauclovian-g:
Question: should the version number be 1.7.0.0 or 1.7.0 or 1.7? Everything else seems to use 4-part version numbers but the older releases are all two or three parts...

langston-barrett (approved):
Seems fine, I'm not exactly sure what to be looking for but nothing sticks out at me.

langston-barrett:
In the absence of actual documentation, let's stay consistent with current practice and use three parts.

what4-transition-system/what4-transition-system.cabal
```diff
@@ -20,7 +20,7 @@ common dependencies
     , lens
     , parameterized-utils  >=2.0  && <2.2
     , text
-    , what4                ^>=1.6
+    , what4                >=1.6 && <1.8
```
RyanGlScott, line 23: We may as well use `what4 ^>=1.7` here, for reasons similar to https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/291#discussion_r2006282229.

(More generally, I think these version bounds aren't buying us much. We are depending on `what4` directly as a subproject in the `cabal.project` file, so whatever version number the `what4` subproject uses will effectively impose a precise version bound on what `what4-transition-system` uses.)

what4/what4.cabal
```diff
@@ -1,6 +1,6 @@
 Cabal-version: 2.4
 Name:          what4
-Version:       1.6.3.0.99
+Version:       1.7.0.0
```
RyanGlScott: The CHANGELOG says `1.7.0`, but this says `1.7.0.0`. We should pick one convention and use it throughout.

(The convention up until this point in `what4`'s history has been to drop the trailing zeroes for release versions—see, for instance, [`what4-1.6`](https://hackage.haskell.org/package/what4-1.6). If we want to stick to that convention, then we should use `1.7` as the next release version.)

what4-transition-system/what4-transition-system.cabal
sauclovian-g: Yeah, probably better to remove it entirely

what4/what4.cabal
sauclovian-g: Must have botched the changelog 😐 I'll set them all to 1.7.

sauclovian-g: FWIW, I think consistency from project to project is less noticeable/important than consistency from release to release in the same project.

what4-transition-system/what4-transition-system.cabal
```diff
@@ -20,7 +20,8 @@ common dependencies
     , lens
     , parameterized-utils  >=2.0  && <2.2
     , text
-    , what4                ^>=1.6
+      -- what4 lives in the same repo; no version bound is needed
+    , what4
```
RyanGlScott, line 24: You might consider factoring this out into its own commit. The same commit could also add this comment to these other `.cabal` files in the same repo:

* https://github.com/GaloisInc/what4/blob/c185a552776d869a4c65b6d1ebc5fa5b5d2eb5af/what4-abc/what4-abc.cabal#L25
* https://github.com/GaloisInc/what4/blob/c185a552776d869a4c65b6d1ebc5fa5b5d2eb5af/what4-blt/what4-blt.cabal#L32

RyanGlScott (approved):

what4-transition-system/what4-transition-system.cabal
sauclovian-g: Good point

sauclovian-g: Done

PR #297, opened 2025-03-27
ci: Move GHC/Cabal compatibility check to improve caching

### LINKED ISSUE #296: ci: GHC/Cabal compatibility check interferes with caching
See https://github.com/GaloisInc/crucible/issues/1349

langston-barrett:
This part of the workflow runs `cabal clean`. However, it was happening after we restored `dist-newstyle` from the cache. I think it was likely deleting some files from the cache, causing worse CI performance than necessary.

Fixes #296.


RyanGlScott (approved):

langston-barrett:
Hmm, this may not be working as I had expected, possibly due to unrelated / existing issues with the caching setup. [This job on this branch](https://github.com/GaloisInc/what4/actions/runs/14114296115/job/39540676838?pr=297) seems to take the same amount of time as [the same job on the main branch](https://github.com/GaloisInc/what4/actions/runs/13983263297/job/39152438344). Strangely, the latter has a cache miss, while the former does not! So how could they take the same amount of time? The whole point of caching is to speed things up!

The `Build` step of both branches begins with logs like
```
In order, the following will be built (use -v for more details):
 - Boolean-0.2.4 (lib:Boolean) (requires download & build)
 - BoundedChan-1.0.3.0 (lib:BoundedChan) (requires download & build)
```
Building dependencies like this should not be necessary when there was a cache hit.

RyanGlScott:
Hm, that suggests that caching is just plain not working. Looking more closely at the YAML configuration, I think I see why:

https://github.com/GaloisInc/what4/blob/4e37eb4abf3db2c52bec09e5612f865a65b7766f/.github/workflows/test.yml#L83-L98

This references `steps.setup-haskell.outputs.cabal-store`, but I don't see an earlier step in the pipeline that defines the `setup-haskell` ID!

PR #301, opened 2025-03-27
ci: Fix caching of Cabal store

### LINKED ISSUE #300: ci: Cache setup refers to nonexistent earlier step
This line refers to `steps.setup-haskell`:

https://github.com/GaloisInc/what4/blob/4e37eb4abf3db2c52bec09e5612f865a65b7766f/.github/workflows/test.yml#L93

but there is no such earlier step! This causes caching to fail.

This problem was exacerbated by https://github.com/GaloisInc/what4/pull/292.

langston-barrett:
Previously, this refered to a nonexistent `step`. Fixes #300.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -89,8 +89,12 @@ jobs:
           # latter as part of the cache key
           key: ${{ env.CACHE_VERSION }}-cabal-${{ matrix.os }}-ghc${{ matrix.ghc }}
         with:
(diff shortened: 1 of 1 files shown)


langston-barrett:
I re-ran this job, and the build step immediately began building local packages: https://github.com/GaloisInc/what4/actions/runs/14115241800/job/39546015962?pr=301

Looks like it works!

RyanGlScott (approved):

PR #312, opened 2025-07-18
Expose uninterpreted functions cache

yav:
When building terms with uninterpreted functions, we'd like to only make a new uninterpreted function once per symbol.
As such, we need to keep track of which symbols we've already processed.   The `ExpreBuilder` has such a cache for when it translates things to uninterpreted functions.  However, for user defined uninterpreted functions we were using a completely separate cache.

I can't think of a good reason to keep these two separate, so this PR exposes the `IORef` for the cache, which makes
it easier to reuse in other contexts that also need to user uninterpreted functions.

To makes things compatable we slightly tweak the type of the cache, but it is essentially the same as it was before.

### DIFF
--- a/what4/src/What4/Expr/Builder.hs
+++ b/what4/src/What4/Expr/Builder.hs
@@ -80,7 +80,7 @@ module What4.Expr.Builder
   , stopCaching
   , exprBuilderSplitConfig
   , exprBuilderFreshConfig
-  , uninterpFnCache
+  , uninterpFnCache, UninterpFunCache
 
     -- * Specialized representations
   , bvUnary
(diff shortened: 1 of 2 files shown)


what4/src/What4/Expr/Builder.hs
```diff
@@ -4639,18 +4643,13 @@ cachedUninterpFn
...
-      -> return fn
-      | otherwise
-      -> fail "Duplicate uninterpreted function declaration."
+  case Map.lookup fn_name fn_cache >>= PM.lookup (arg_types Ctx.:> ret_type) of
+    Just (SymFnWrapper fn) -> pure fn
     Nothing -> do
       fn <- handler sym fn_name arg_types ret_type
-      atomicModifyIORef' (sbUninterpFnCache sym) (\m -> (Map.insert fn_key (SomeSymFn fn) m, ()))
+      _ <- atomicModifyIORef' (sbUninterpFnCache sym)
+        (\m -> (Map.alter (Just . PM.insert (arg_types Ctx.:> ret_type) (SymFnWrapper fn) . fromMaybe PM.empty) fn_name m, ()))
```
RyanGlScott, line 4651: Worth factoring out the higher-function that you pass to `Map.alter` into its own function, I'd say—this line is pretty long.

RyanGlScott (approved):
Can you mention this change in the changelog?

PR #313, opened 2025-08-11
Support GHC 9.12

### LINKED ISSUE #286: Support GHC 9.12
https://www.haskell.org/ghc/download_ghc_9_12_1.html

langston-barrett:
Fixes #286.


RyanGlScott:
The bulk of this PR looks reasonable to me, although I will note that neither `what4-abc` nor `what4-blt` will build at the moment on sufficiently recent `cabal` versions due to https://github.com/GaloisInc/abcBridge/issues/23 and https://github.com/GaloisInc/blt/issues/5, respectively. It might be nice to fix them and then bump the `abcBridge`/`blt` submodules accordingly.

langston-barrett:
Bumped the submodules.

RyanGlScott:
Blocked on https://github.com/kquick/kvitable/issues/7 (which, in turn, is blocked on https://github.com/kquick/sayable/pull/3)

cabal.project
```diff
@@ -14,3 +14,8 @@ optional-packages:
   dependencies/abcBridge/
   dependencies/aig/
   dependencies/blt/
+
+-- See https://github.com/kquick/kvitable/issues/7
+allow-newer: kvitable:base
+-- See https://github.com/GaloisInc/lumberjack/issues/9
+allow-newer: lumberjack:base
```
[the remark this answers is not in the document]
RyanGlScott: I've removed these in https://github.com/GaloisInc/what4/pull/313/commits/332dc1e900cc4007bec029f3e7187359927a55c2.

RyanGlScott:
I've pushed some commits (which we may want to squash before landing) that should address the remaining concerns.

RyanGlScott:
> There's an unrelated change in ABC.hs

Unfortunately, this change is actually necessary to make `what4-abc` compile with GHC 9.12. See https://gitlab.haskell.org/ghc/ghc/-/issues/27149.

what4-abc/src/What4/Solver/ABC.hs
```diff
@@ -573,7 +573,7 @@ evalNonce ntk n eval_fn fallback = do
   -- Look to see if literals have been assigned to expression.
   mnm <- liftST $ H.lookup (nameCache ntk) n
   case mnm of
-    Just (B l) -> return $ eval_fn l
+    Just (B l) -> return (eval_fn l)
```
kquick, line 578: Perhaps we should add a comment to this line:
```... -- do not switch to using $ while GHC 9.12 is supported, see https://gitlab.haskell.org/ghc/ghc/-/issues/27149```

I know that seems like a lot of comment for a little line, but otherwise this change looked cosmetic and I could see it easily being accidentally changed back to the original form, and having the  bug reference would be helpful (and perhaps for other occurrences as well).

RyanGlScott: Done.

sauclovian-g:
what4-transition-system needs further changes, FWIW. I have a branch, I'm waiting to get a successful SAW build in case there's further issues.

RyanGlScott:
> what4-transition-system needs further changes

Can you elaborate on what you mean here? I am able to build `what4-transition-system` using GHC 9.12 without any issues or build warnings, so I must be missing something.

sauclovian-g:
Maybe it only comes up when building it with SAW; however, there it needs to allow containers 0.7 and ansi-wl-pprint 1.0.x. The latter is probably some interaction with SAW; the former though I thought ships with ghc...

RyanGlScott:
Good catch, I was admittedly not very thorough in checking for outdated upper version bounds.

sauclovian-g:
Apparently it does only come up with SAW; I successfully built language-sally against it without those changes. Not sure what's going on there, but not super inclined to waste time worrying about it...

PR #316, opened 2025-08-29
Make `sbvIntTerm` work properly for length-1 bitvectors

### LINKED ISSUE #315: Unexpected `arithmetic underflow` exception when checking satisfiability involving `BaseBVType 1`
In Cryptol, I noticed that the following property could be checked for satisfiability using an SBV-based prover:

```
Cryptol> :set prover=sbv-z3
Cryptol> :sat \(x : [1]) -> toSignedInteger x == toInteger x
Satisfiable
(\(x : [1]) -> toSignedInteger x == toInteger x) 0x0 = True
(Total Elapsed Time: 0.013s, using "Z3")
```

On the other hand, checking the same property using a What4-based prover would cause an unexpected exception to be thrown:

```
Cryptol> :set prover=w4-z3
Cryptol> :sat \(x : [1]) -> toSignedInteger x == toInteger x

What4 exception:
arithmetic underflow
```

I believe this to be a bug in the `what4` library itself, as you can trigger the same exception using nothing but the `what4` API:

```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
module Main where

import Data.Foldable (forM_)

import What4.Expr
         ( Expr, ExprBuilder, FloatModeRepr(..), newExprBuilder
         , BoolExpr, groundEval
         , EmptyExprBuilderState(..) )
import What4.Interface
         ( BaseBVType, BaseTypeRepr(..), IsExprBuilder(..)
         , IsSymExprBuilder(..), knownNat, safeSymbol )
import What4.Solver
         (defaultLogData, z3Options, withZ3, SatResult(..))
import What4.Protocol.SMTLib2
         (assume, sessionWriter, runCheckSat)


z3executable :: FilePath
z3executable = "z3"

main :: IO ()
main = do
  extendConfig z3Options (getConfiguration sym)

  x <- freshConstant sym (safeSymbol "x") (BaseBVRepr (knownNat @1))
  xsi <- sbvToInteger sym x
  xi <- bvToInteger sym x
  eq <- intEq sym xsi xi
  checkModel sym eq [("x", x)]

-- | Determine whether a predicate is satisfiable, and print out the values of a
-- set of expressions if a satisfying instance is found.
checkModel ::
  ExprBuilder t st fs ->
  BoolExpr t ->
  [(String, Expr t (BaseBVType w))] ->
  IO ()
checkModel sym f es = do
  -- We will use z3 to determine if f is satisfiable.
  withZ3 sym z3executable defaultLogData $ \session -> do
    -- Assume f is true.
    assume (sessionWriter session) f
    runCheckSat session $ \result ->
      case result of
        Sat (ge, _) -> do
          forM_ es $ \(nm, e) -> do
            v <- groundEval ge e
```
```
λ> main
*** Exception: arithmetic underflow
```

Note that this does not happen if you use `BaseBVRepr (knownNat @n)` where `n >= 2`.

RyanGlScott:
Fixes #315.

### DIFF
--- a/what4/src/What4/Protocol/SMTWriter.hs
+++ b/what4/src/What4/Protocol/SMTWriter.hs
@@ -1598,11 +1598,11 @@ bvIntTerm :: forall v w
           => NatRepr w
           -> v
           -> v
-bvIntTerm w x = sumExpr powersOfTwo
+bvIntTerm w x = sumExpr digits
  where -- Precondition: 1 <= w. This is upheld by the `1 <= w` constraint in
        -- bvIntTerm's type signature.
-       powersOfTwo :: [v]
-       powersOfTwo = (\i -> digit (i-1)) <$> [1..natValue w]
+       digits :: [v]
+       digits = (\i -> digit (i-1)) <$> [1..natValue w]
 
        digit :: Natural -> v
        digit d = ite (bvTestBit w d x)
@@ -1618,7 +1618,7 @@ sbvIntTerm :: forall v w
            => NatRepr w
            -> v
            -> v
-sbvIntTerm w x = sumExpr (signedOffset : powersOfTwo)
+sbvIntTerm w x = sumExpr (signedOffset : digits)
  where -- Precondition: 1 <= w. This is upheld by the `1 <= w` constraint in
        -- sbvIntTerm's type signature.
        signedOffset :: v
@@ -1626,8 +1626,8 @@ sbvIntTerm w x = sumExpr (signedOffset : powersOfTwo)
                           (fromInteger (negate (2^(widthVal w - 1))))
                           0
 
-       powersOfTwo :: [v]
-       powersOfTwo = (\i -> digit (i-1)) <$> [1..natValue w]
+       digits :: [v]
+       digits = (\i -> digit (i-1)) <$> [1..natValue w]
 
        digit :: SupportTermOps v => Natural -> v
        digit d = ite (bvTestBit w d x)


what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -1604,17 +1604,20 @@ bvIntTerm w x = sumExpr ((\i -> digit (i-1)) <$> [1..natValue w])
                      (fromInteger (2^d))
                      0
 
-sbvIntTerm :: SupportTermOps v
+sbvIntTerm :: (SupportTermOps v, 1 <= w)
            => NatRepr w
            -> v
            -> v
-sbvIntTerm w0 x0 = sumExpr (signed_offset : go w0 x0 (natValue w0 - 2))
+sbvIntTerm w0 x0 = sumExpr (signed_offset : go w0 x0 (intValue w0 - 2))
```
kquick, line 1611: This took a little bit of staring to realize the main problem is this line and not the one with the explaining comment below.  This works, although I still worry that the super-degenerate case of a length-0 bitvector would trigger an underflow exception in `signed_offset`.   An alternative could be:
```
sbvIntTerm w0 _ | natValue w0 < 2 = sumExpr [0]
sbvIntTerm w0 x0 = ... original unchanged and dropping the `otherwise` clause below ...
```
although I'm not sure about the `sumExpr` arg for the length-1 bitvector case.

kquick: Although I had seen the constraint you added on `w`, I was worried that it could be called with a `v` that was a different size than was represented by `w`.  However, some light checking indicates that this is unlikely and would be very difficult to do (if even possible), so as a result I'm less concerned about this issue.  Adjusting the comment location would be appreciated though.

kquick (approved):

.github/workflows/test.yml
```diff
@@ -37,17 +37,17 @@ jobs:
           submodules: true
 
       - name: Install Nix
-        uses: cachix/install-nix-action@v16
+        uses: cachix/install-nix-action@v31
```
langston-barrett, line 40: Unrelated to the bulk of this PR, but I wonder if we might want to start using Dependabot on this repo too so that we don't get so far out-of-date, and so that we don't have to manually bump these numbers. IMO it's been working fine in p-u since https://github.com/GaloisInc/parameterized-utils/pull/195, see e.g., https://github.com/GaloisInc/parameterized-utils/pulls?q=is%3Apr+is%3Aclosed+author%3Aapp%2Fdependabot.

langston-barrett (approved):

what4/src/What4/Protocol/SMTWriter.hs
RyanGlScott: The longer that I stare at the implementation of `sbvIntTerm`, the more I realize how unnecessarily complicated it is. It's doing basically the same thing as `bvIntTerm`, but adding an additional summand to make it negative if need be. That's it!

I've pushed an additional commit that refactors the code to make the similarities between `bvIntTerm` and `sbvIntTerm` more obvious. As a bonus, this makes it much clearer where the potential underflow issue would be (in the newly christened `powersOfTwo` helper functions). Let me know if you like this version better.

PR #323, opened 2025-09-30
Replace `lens` with `microlens{,-mtl,-th}`

langston-barrett:
The motivation is that `lens` has many, many (transitive) dependencies. This leads to:

- Slow build times (the `microlens` README claims that it compiles in 3.5s whereas `lens` takes  4m10s)
- Longer GHC upgrade latency (we have to wait until all of `lens`'s dependencies support a new GHC version)
- Smaller GHC support windows (our GHC support window is limited by that of all of our transitive dependencies)
- Supply chain risks such as lack of maintenance and even supply chain attacks (see recent NPM attacks)

This currently depends on https://github.com/GaloisInc/parameterized-utils/pull/208, it should wait until that PR is merged and released.


RyanGlScott (approved):

what4/src/What4/Config.hs
```diff
@@ -318,8 +317,11 @@ data OptionSetting (tp :: BaseType) =
 
 instance Show (OptionSetting tp) where
   show = (<> " option setting") .
-         LC.cons '\'' . flip LC.snoc '\'' .
+         cons '\'' . flip snoc '\'' .
```
kquick, line 320: I don't have any strong objections to this, but it seems like this line could just be `("'" <>) . (<> "'")` with no need for auxiliary functions.

kquick (approved):

what4/src/What4/Config.hs
langston-barrett: Agreed, I'll keep this as-is for now just to keep these changes minimal in case there are subtle issues to debug.

PR #343, opened 2026-02-03
w4smt2: A bad SMT solver

langston-barrett:
See the README.

Created with Claude code. Like the README says, the implementation is not super important (though I have tried to make it reasonably readable). The quasiquoter in particular is entirely vibe-coded. I personally think that this is OK given the intent of the tool, but would be happy to take a pass to try to improve it if desired.

### DIFF
--- a/w4smt2/test/Main.hs
+++ b/w4smt2/test/Main.hs
@@ -21,8 +21,9 @@ import Test.Tasty.Golden (goldenVsString)
 import Test.Tasty.HUnit (testCase, (@?=))
 import What4.Expr (newExprBuilder, EmptyExprBuilderState(EmptyExprBuilderState))
 import What4.FloatMode (FloatModeRepr(FloatUninterpretedRepr))
-import What4.SatResult (SatResult(Sat, Unsat, Unknown))
+import What4.SatResult (SatResult(Sat, Unsat))
 
+import W4SMT2.Exec (erResults)
 import W4SMT2.Solve (solve)
 import W4SMT2.Parser qualified as Parser
 import W4SMT2.Pretty qualified as Pretty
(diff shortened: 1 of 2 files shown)


what4/src/What4/Expr/App.hs
```diff
@@ -788,6 +788,12 @@ instance FoldableFC App where
 traverseApp :: (Applicative m, OrdF f, Eq (f (BaseBoolType)), HashableF f, HasAbsValue f)
             => (forall tp. e tp -> m (f tp))
             -> App e utp -> m ((App f) utp)
+
+-- This is the type at its use in 'evalBoundVars'
+{-# SPECIALIZE traverseApp ::
+  (forall tp. Expr t tp -> IO (Expr t tp))
+  -> App (Expr t) utp -> IO ((App (Expr t)) utp) #-}
```
RyanGlScott: Redundant parentheses:


```suggestion
  -> App (Expr t) utp -> IO (App (Expr t) utp) #-}
```

langston-barrett: Fixed, also fixed in the actual type signature

w4smt2/src/W4SMT2/SomeExpr.hs
```diff
@@ -0,0 +1,22 @@
...
+import Unsafe.Coerce (unsafeCoerce)
+
+import What4.Interface (SymExpr)
+
+newtype SomeExpr sym = UnsafeSome (SymExpr sym Any)
+
+{-# COMPLETE SomeExpr #-}
+pattern SomeExpr :: forall sym a. SymExpr sym a -> SomeExpr sym
+pattern SomeExpr x <- UnsafeSome ((unsafeCoerce :: SymExpr sym Any -> SymExpr sym a) -> x)
+  where SomeExpr x = UnsafeSome ((unsafeCoerce :: SymExpr sym a -> SymExpr sym Any) x)
```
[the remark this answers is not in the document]
langston-barrett: I think I'll keep it for now, depending on another package seems like overkill and I would like to avoid overhead.

RyanGlScott (approved):
LGTM, but please do file a follow-up issue about refactoring the parsing-related code.

PR #351, opened 2026-02-11
who2: Towards a faster expression builder

langston-barrett:
## Motivation

When working on `w4smt2`, I noticed that `ExprBuilder` added considerable overhead to SMT queries, and the overhead grew with larger queries. Profiling revealed that the culprit was What4's normalizing datastructures such as `WeightedSum`, which have `O(n log n)` worst-case construction time. I became concerned that our higher-level tools might be paying an excessive performance penalty by going through What4.

## Approach

Who2 provides an alternative expression builder that performs local rewrites and tracks abstract domains, but uses only approximately linear time by default.

Many of Who2's performance-oriented features (e.g., hash-consing) can be controlled via Cabal flags, allowing consumers to try out different features depending on their use-case. When disabled, they result in clearly dead branches that are pruned by the compiler.

## Results

Benchmarks show that Who2 offers extremely low overhead but still manages to solve a substantial number of problems without consulting a solver. It would certainly be worthwhile to consider benchmarking if Who2 speeds up our actual workloads from Crucible, SAW, and GREASE.

## Testing

Since the correctness of Who2 is crucial to any tools that might be built on it (and validity of benchmarking results), I have put special care into fairly extensive testing.

- I have run `who2` via `w2smt2` on many SMT-Lib sample problems and verified its agreement with SMT solvers and What4.
- Each rewrite has a corresponding unit test, and that statement is itself tested (using `-- test: ` comments in the source). Each rewrite test is itself validated against Z3.
- Each rewrite has a corresponding Cryptol `:prove` command establishing its validity, if Cryptol is available these are proved in the test suite.
- Each constructor has a golden SMT-Lib output test which is validated by Z3, and that statement is itself tested (using `-- test-smt:` comments in the source).
- I created a Hedgehog generator for a type `ExprBuilderAPI`. This is like the existing "template" tests.
  - There is a Hedgehog test stating that `Builder` reduces expressions without variables to literals.
  - There is a Hedgehog test stating that `Builder` never builds empty datastructures (e.g., semiring sums)
  - There is a Hedgehog test stating that `ExprBuilderAPI`, as interpreted by `Builder`, can generate every constructor of `Builder`'s `Expr`. This ensures good coverage in the generator.
  - I created a `TestBuilder` that uses the same `SymExpr` type as Who2's `Builder`, but without rewriting and with all-top abstract domains. There is a Hedgehog test stating that these builders produce equisatisfiable predicates. I did some manual mutation testing (changing some rewrites) and confirmed that it found the issues I introduced.
- There are property-based tests for laws of typeclass `instance`s in the code.

There are, in fact, more test modules than source modules.

## Documentation

See especially the Haddocks on:

- `Who2.Builder`
- `Who2.Config`
- `Who2.Expr.App`

## Status and future work

Only `QF_BV` operations are implemented at the moment.

TODO (before merging):

- [x] ~~Represent `or` as negated `and` a la What4~~
- [x] Deduplicate a bit of logic between rewrites (re: `PolarizedBloomSeq`)
- [x] Rebase after merging #343 
- [x] Create a pre-hashed `Seq` supporting appending
- [x] Generalize `BloomSeq` to a map-like datastructure, use for a more comprehensive `WeightedSum`-like datastructure
- [x] Overview of datastructures
- [x] Optional hash-consing
- [x] Optional AST normalization
- [x] Optional patricia-tree based datastructures
- [x] README with implementation status and next steps, especially solver interaction
- [x] Audit typeclass instances (especially horrible `TemplateHaskell` instances)
- [x] Test for generator completeness

Possible extensions:

- Improve abstract domains
- Support other SMT-Lib logics (integers, reals)
- What4-like interface to solvers (currently `w2smt2` just serializes to file)

I did use Claude code in building who2.


w4smt2/bench/Benchmark/Config.hs
```diff
@@ -90,7 +97,7 @@ solverOption = Opt.option (Opt.str >>= parseSolver)
   ( Opt.long "solver"
   <> Opt.short 's'
   <> Opt.metavar "SOLVER"
-  <> Opt.help "Solver to run (can be specified multiple times). Valid: z3, yices, cvc5, bitwuzla, w4, w4z3, w4yices, w4cvc5, w4bitwuzla"
+  <> Opt.help "Solver to run (can be specified multiple times). Valid: z3, yices, cvc5, bitwuzla, w4, w4z3, w4yices, w4cvc5, w4bitwuzla, w2, w2z3, w2yices, w2cvc5, w2bitwuzla"
```
RyanGlScott, line 100: This line is becoming rather large. I wonder if we should generate this line programmatically by enumerating over all of the different `Solver` constructors.

w4smt2/bench/Benchmark/Runner.hs
```diff
@@ -85,8 +86,21 @@ data RunningProcess = RunningProcess
 
 buildW4SMT2 :: IO FilePath
 buildW4SMT2 = do
-  _ <- readCreateProcess (Proc.proc "cabal" ["build", "-O2", "exe:w4smt2"])
-  path <- readCreateProcess (Proc.proc "cabal" ["list-bin", "-O2", "exe:w4smt2"])
+  _ <- readCreateProcess (Proc.proc "cabal" ["build", "-O2", "--ghc-options=-O2", "exe:w4smt2"])
+  path <- readCreateProcess (Proc.proc "cabal" ["list-bin", "-O2", "--ghc-options=-O2", "exe:w4smt2"])
```
RyanGlScott, line 90: Why do we need to pass both `-O2` and `--ghc-options=-O2` here?

w4smt2/src/W4SMT2/Parser.hs
```diff
@@ -211,57 +210,65 @@ parseAtomLiteral ::
...
+      let !width = fromIntegral @Int @Natural (Text.length hexDigits * 4)
+      case NatRepr.mkNatRepr width of
+        Some w -> case NatRepr.testLeq (NatRepr.knownNat @1) w of
+          Just NatRepr.LeqProof -> do
+            bv <- WI.bvLit sym w (BV.mkBV w val)
+            return (Just (SomeExpr bv))
+          Nothing -> return Nothing
+  | otherwise = return Nothing
+  where
+    isHexDigit c = (c >= '0' && c <= '9') || (c >= 'a' && c <= 'f') || (c >= 'A' && c <= 'F')
```
RyanGlScott, line 263: If I'm not mistaken, this is just re-implementing [`isHexDigit`](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Data-Char.html#v:isHexDigit) from `Data.Char`, right? Why not just use that?

who2/src/Who2/Expr/InstanceHelpers.hs
```diff
@@ -0,0 +1,60 @@
...
+import qualified Data.Parameterized.Classes as PC
+
+-- | Implement TestEquality via Eq
+viaEq ::
+  Eq a =>
+  a ->
+  a ->
+  Maybe (b :~: b)
+viaEq x y = if x == y then Just Refl else Nothing
+{-# INLINE viaEq #-}
```
RyanGlScott, line 25: I find this (as well as `viaEqBy`, `viaOrd`, and `viaOrdBy` below) to be very strange definitions. What is the value of producing a `b :~: b` witness when you conclude that two values of type `a` (which are unrelated types) are equal?

who2/README.md
```diff
@@ -0,0 +1,20 @@
+# Who2
+
+Who2 provides a configurable alternative to What4's `ExprBuilder` that uses only
+approximately linear time by default. See the Haddocks in the following modules
+for more information:
```
RyanGlScott, line 5: I'd appreciate some commentary here about how `who2` is tested and benchmarked.

what4/src/What4/Protocol/SExp.hs
```diff
@@ -46,9 +45,9 @@ parseNextWord = do
   skipSpaceOrNewline
   mappend (takeWhile1 isAlphaNum) (fail "Unexpected end of stream.")
 
-data SExp = SAtom Text
-          | SString Text
-          | SApp [SExp]
+data SExp = SAtom !Text
+          | SString !Text
+          | SApp ![SExp]
```
RyanGlScott, line 50: I'm a bit surprised to see changes to `what4` itself as part of this PR. What motivated them?

who2/src/Who2/Config.hs
```diff
@@ -0,0 +1,236 @@
...
+--
+-- To enable/disable features at build time:
+-- @
+-- cabal build -f+hash-consing          # Enable hash-consing (default: enabled)
+-- cabal build -f-bloom-filter          # Disable bloom filtering
+-- cabal build -f+fancy-hash            # Enable polynomial rolling hash
+-- @
+--
+-- All constants use @INLINE@ pragmas for zero-overhead abstraction - when
+-- disabled, the compiler eliminates dead code paths entirely.
```
RyanGlScott, line 17: For what it's worth, I would be very surprised if these `INLINE` pragmas are actually necessary. The definitions of these `Bool` values are so simple that GHC will almost assuredly inline them anyway.

who2/test/Who2/Cryptol.hs
```diff
@@ -0,0 +1,138 @@
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | Tests for Cryptol proofs embedded in source code comments.
+--
+-- Searches for comments like @-- Cryptol> :prove \\(x : [8]) -> x + 0 == x@
```
RyanGlScott, line 6: As far as I can tell, these Cryptol tests are not tested in the CI. I worry about the potential for this to bitrot quickly if we don't.

RyanGlScott (commented):
This adds over 14,000 lines of code, and as such, I don't have time to review every line of code individually. Given that this is an opt-in component that we are not using for foundational correctness yet, I am not worried about this too much at the moment. That being said, if we decide to explore using this in tools like Crucible or SAW, I would worry about using this unless we had someone audit the code more closely.

I have left a variety of comments based on very quick scans of the code.

langston-barrett:
I don't think the perf numbers are compelling enough to keep this around.

PR #356, opened 2026-03-17
Fix side conditions lost after `reset-assertions`

langston-barrett:
After `reset` (which sends `(reset-assertions)` or equivalent), side conditions added by `addPartialSideCond` (e.g., Nat >= 0 for fresh Nat variables) are cleared from the solver. However, the `WriterConn` cache still has `DeleteNever` entries for these variables, so `mkExpr` doesn't re-add the side conditions when the variables are next encountered.

The test creates a `Nat` variable `n` (with side condition `n >= 0`), links it to an unconstrained integer `m` via `m = n`, resets the solver, then checks that `m < 0` is still UNSAT. Without the fix, the side condition is lost and `m < 0` becomes SAT.

The fix adds a `sideConditions` field to `WriterConn` that accumulates side condition terms as they are asserted via `runOnLiveConnection`. After `reset` sends `(reset-assertions)`, `reassertSideConditions` replays all accumulated side conditions to restore the solver's invariants.

### DIFF
--- a/what4/src/What4/Protocol/SMTWriter.hs
+++ b/what4/src/What4/Protocol/SMTWriter.hs
@@ -64,6 +64,7 @@ module What4.Protocol.SMTWriter
               )
   , connState
   , newWriterConn
+  , reassertSideConditions
   , resetEntryStack
   , popEntryStackToTop
   , entryStackHeight
@@ -643,6 +644,12 @@ data WriterConn t (h :: Type) =
              , consumeAcknowledgement :: AcknowledgementAction t h
                -- ^ Consume an acknowledgement notifications the solver, if
                --   it produces one
+             , sideConditions :: !(IORef [Term h])
+               -- ^ Side conditions that must be re-asserted after
+               --   a @(reset-assertions)@ command. These arise from
+               --   'addPartialSideCond' for variables cached with
+               --   'DeleteNever', whose cache entries survive a reset but whose
+               --   solver-level assertions do not.
              }
 
 -- | An action for consuming an acknowledgement message from the solver,
@@ -729,6 +736,7 @@ newWriterConn h in_h ack solver_name beStrict features bindings cs = do
   entry <- newStackEntry
   stk_ref <- newIORef [entry]
   r <- newIORef emptyState
+  scRef <- newIORef []
   return $! WriterConn { smtWriterName = solver_name
                        , connHandle    = h
                        , connInputHandle = in_h
@@ -742,6 +750,7 @@ newWriterConn h in_h ack solver_name beStrict features bindings cs = do
                        , varBindings  = bindings
                        , connState    = cs
                        , consumeAcknowledgement = ack
+                       , sideConditions = scRef
                        }
 
 -- | Strictness level for parsing solver responses.
@@ -1075,6 +1084,16 @@ bindVarAsFree conn var = do
       addCommand conn $ declareCommand conn var_name Ctx.empty smt_type
       cacheValueExpr conn (bvarId var) DeleteOnPop $ SMTName smt_type var_name
 
+-- | Re-assert all side conditions that were recorded by 'addPartialSideCond'.
+--
+-- See 'sideConditions' for more details.
+reassertSideConditions :: SMTWriter h => WriterConn t h -> IO ()
+reassertSideConditions conn = do
+  conds <- readIORef (sideConditions conn)
+  -- This adds them back in the reverse order they were originally added, but
+  -- that shouldn't matter semantically.
+  mapM_ (assumeFormula conn) conds
+
 -- | Assume that the given formula holds.
 assumeFormula :: SMTWriter h => WriterConn t h -> Term h -> IO ()
 assumeFormula c p = addCommand c (assertCommand c p)
@@ -1358,6 +1377,21 @@ addSideCondition nm t = do
      fail $ "Cannot add a side condition within a function needed to define the "
        ++ nm ++ " term created at " ++ show loc ++ "."
 
+-- | Add a persistent side condition that will be re-asserted after reset.
+--
+-- This is specifically for side conditions on DeleteNever variables (like
+-- UninterpVarKind variables from freshNat, freshConstant, etc.) whose
+-- constraints must persist across reset.
+addPersistentSideCondition ::
+   SMTWriter h =>
+   WriterConn t h ->
+   String {- ^ Reason that condition is being added. -} ->
+   Term h {- ^ Predicate that should hold. -} ->
+   SMTCollector t h ()
+addPersistentSideCondition conn nm t = do
+  addSideCondition nm t
+  liftIO $ modifyIORef' (sideConditions conn) (t:)
+
 addPartialSideCond ::
   forall t h tp.
   SMTWriter h =>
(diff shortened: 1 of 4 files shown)


langston-barrett:
TBQH, I don't deeply understand the area being fixed here, this bug was identified and fixed by an agent. But the test seems compelling, and seems to match the fix.

RyanGlScott (commented):
If I understand the chain of thought here:

1. `(reset-assumptions)` normally resets _all_ of the assumptions from the assumption stack.
2. Nevertheless, there are things that we sometimes want to persist even after a `(reset-assumptions)` call. The side condition `n >= 0` that arises from creating a fresh `Nat` `n` is one example of this, as if we drop this side condition, then it is possible to unsoundly conclude that `n` negative after a call to `(reset-assumptions)`. Let's call these "persistent assertions".
3. This PR claims to achieve that by changing the treatment of `DeleteNever`-tagged assertions such that they are re-asserted after `(reset-assumptions)`.

Questions that I would like to answer before going forward with this:

* How exactly does `freshNat` make use of `DeleteNever` under the hood?
* How can we be sure that this only re-asserts persistent assertions and doesn't re-assert non-persistent ones? For instance, I note that this adds a persistent assertion (to the `sideConditions` `IORef`) every time that `recordSideCondFn` is invoked---how do we know that that always corresponds to a persistent assertion?

langston-barrett:
Yes, that's correct.

> How exactly does freshNat make use of DeleteNever under the hood?

From an LLM (I checked this out, it's true):

> The chain of execution is:
> 
> 1. **`freshNat`** → **`freshBoundedNat`** → **`freshBoundedInt`**
>    - Location: `what4/src/What4/Interface.hs:572-573, 566, 4458-4462`
> 
> 2. **`freshBoundedInt`** creates a `BoundVarExpr` with:
>    - Type: `BaseIntegerRepr`
>    - Kind: `UninterpVarKind` (uninterpreted variable)
>    - Abstract value: `Just (MultiRange (Inclusive 0) Unbounded)` representing the constraint `n >= 0`
> 
> 3. When this expression is **first sent to the solver** (via `mkFormula`/`mkSMTTerm`), the `mkExpr` function processes the `BoundVarExpr`.
> 
> 4. For `UninterpVarKind` variables (`what4/src/What4/Protocol/SMTWriter.hs:1944-1961`):
>    ```haskell
>    UninterpVarKind -> do
>      conn <- asks scConn
>      cacheWriterResult (bvarId var) DeleteNever $ do  -- Line 1946: Cache with DeleteNever
>        checkVarTypeSupport var
>        var_name <- liftIO $ getSymbolName conn (VarSymbolBinding var)
>        smt_type <- getBaseSMT_Type var
> 
>        liftIO $ do
>          declareTypes conn smt_type
>          addCommand conn $ declareCommand conn var_name Ctx.empty smt_type
> 
>        -- Add assertion based on var type (Line 1958)
>        addPartialSideCond conn (fromText var_name) smt_type (bvarAbstractValue var)
> 
>        return $ SMTName smt_type var_name
>    ```
> 
> **Key insight**: `freshNat` doesn't directly use `DeleteNever`. Rather, when an `UninterpVarKind` bound variable (which `freshNat` creates) is sent to the solver, the SMT writer automatically:
> - Caches it with `DeleteNever` (line 1946)
> - Declares it to the solver
> - Calls `addPartialSideCond` with the abstract value, generating the `n >= 0` constraint (line 1958)
> 
> The `DeleteNever` lifetime ensures that the variable remains in the cache even after `(reset-assertions)`, so it won't be re-declared on subsequent uses. However, before this PR, the side condition `n >= 0` would be lost after reset.

For your second question:

> How can we be sure that this only re-asserts persistent assertions and doesn't re-assert non-persistent ones? For instance, I note that this adds a persistent assertion (to the sideConditions IORef) every time that recordSideCondFn is invoked---how do we know that that always corresponds to a persistent assertion?

Good question! In fact, there were unnecessary side conditions being added for operations like `sqrt`. I fixed this and added a test showing that it doesn't happen. I also added a test verifying nominal behavior of reset.

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -643,6 +644,12 @@ data WriterConn t (h :: Type) =
              , consumeAcknowledgement :: AcknowledgementAction t h
                -- ^ Consume an acknowledgement notifications the solver, if
                --   it produces one
+             , sideConditions :: !(IORef [Term h])
+               -- ^ Side conditions that must be re-asserted after
+               --   a @(reset-assertions)@ command. These arise from
+               --   'addPartialSideCond' for variables cached with
+               --   'DeleteNever', whose cache entries survive a reset but whose
+               --   solver-level assertions do not.
```
RyanGlScott: Perhaps this should be called "`persistentSideConditions`"? Especially since we now have separate `addSideCondition` and `addPersistentSideCondition` functions.

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -1075,6 +1084,16 @@ bindVarAsFree conn var = do
       addCommand conn $ declareCommand conn var_name Ctx.empty smt_type
       cacheValueExpr conn (bvarId var) DeleteOnPop $ SMTName smt_type var_name
 
+-- | Re-assert all side conditions that were recorded by 'addPartialSideCond'.
+--
+-- See 'sideConditions' for more details.
+reassertSideConditions :: SMTWriter h => WriterConn t h -> IO ()
+reassertSideConditions conn = do
```
RyanGlScott: Similarly, I think it might be worth calling this "`reassertPersistentSideConditions`" to make the nomenclature consistent.

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -1075,6 +1084,16 @@ bindVarAsFree conn var = do
       addCommand conn $ declareCommand conn var_name Ctx.empty smt_type
       cacheValueExpr conn (bvarId var) DeleteOnPop $ SMTName smt_type var_name
 
+-- | Re-assert all side conditions that were recorded by 'addPartialSideCond'.
```
RyanGlScott: Note very side condition that `addPartialSideCond` adds is a persistent one, correct? It might be worth clarifying that this only re-asserts the persistent ones.

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -1358,6 +1377,21 @@ addSideCondition nm t = do
...
+   SMTWriter h =>
+   WriterConn t h ->
+   String {- ^ Reason that condition is being added. -} ->
+   Term h {- ^ Predicate that should hold. -} ->
+   SMTCollector t h ()
+addPersistentSideCondition conn nm t = do
+  addSideCondition nm t
+  liftIO $ modifyIORef' (sideConditions conn) (t:)
+
 addPartialSideCond ::
```
RyanGlScott, line 1401: This is a pretty key function in establishing persistent side conditions, and I think it would be worth adding some Haddocks to elaborate what it does.

what4/test/OnlineSolverTest.hs
```diff
@@ -293,6 +293,150 @@ longTimeTest (SolverName nm, AnOnlineSolver (Proxy :: Proxy s), features, opts,
 ----------------------------------------------------------------------
 
 
+-- | Test that side conditions (e.g., Nat >= 0) are preserved after 'reset'.
```
RyanGlScott: Clarify:

```suggestion
-- | Test that persistent side conditions (e.g., Nat >= 0) are preserved after 'reset'.
```

Other parts of the comments in this module might also benefit from similar changes.

what4/test/OnlineSolverTest.hs
```diff
@@ -293,6 +293,150 @@ longTimeTest (SolverName nm, AnOnlineSolver (Proxy :: Proxy s), features, opts,
...
+-- declares it and adds a side condition @n >= 0@ via 'addPartialSideCond'.
+-- After 'reset' (which sends @(reset-assertions)@), these side conditions
+-- should still hold.
+--
+-- We use an unconstrained integer @m@ linked to @n@ via @m = n@ to prevent
+-- the ExprBuilder from optimizing away the check based on abstract domains.
+mkResetSideCondTest :: (SolverTestData, SolverVersion) -> TestTree
+mkResetSideCondTest ((SolverName nm, AnOnlineSolver (Proxy :: Proxy s), features, opts, _), _)
+  | nm `elem` ["Bitwuzla", "Boolector", "STP"]
+  = testCase nm $ assertBool "skipped (no integer support)" True
```
RyanGlScott: I'm not fond of hard-coding the names of individual solvers here, especially since the actual property of interest is whether a solver supports integer arithmetic or not. I'd favor guarding this on something like `` if supportedFeatures conn `hasProblemFeature` useIntegerArithmetic `` instead.

(Similarly for the other tests below.)

RyanGlScott (commented):
Thanks, I like this version of the PR much better. (I am being extra cautious here since introducing a bug in the way assertions are handled could potentially affect soundness.)

Please make sure to mention this in the changelog.

langston-barrett:
It appears that the tests are failing on the "STP" prover in CI, likely because I don't have that one installed. Will fix.

langston-barrett:
I just went ahead and disabled them for STP specifically.

what4/test/OnlineSolverTest.hs
```diff
@@ -293,6 +293,150 @@ longTimeTest (SolverName nm, AnOnlineSolver (Proxy :: Proxy s), features, opts,
...
+-- When a fresh Nat variable is created and sent to the solver, 'mkExpr'
+-- declares it and adds a side condition @n >= 0@ via 'addPartialSideCond'.
+-- After 'reset' (which sends @(reset-assertions)@), these side conditions
+-- should still hold.
+--
+-- We use an unconstrained integer @m@ linked to @n@ via @m = n@ to prevent
+-- the ExprBuilder from optimizing away the check based on abstract domains.
+mkResetSideCondTest :: (SolverTestData, SolverVersion) -> TestTree
+mkResetSideCondTest ((SolverName nm, AnOnlineSolver (Proxy :: Proxy s), features, opts, _), _)
+  | not (hasProblemFeature features useIntegerArithmetic) || nm == "STP"
```
RyanGlScott: Please leave a comment here (and elsewhere) saying that we disable STP since it times out on this particular problem.

RyanGlScott (approved):

PR #359, opened 2026-04-09
`what4-transition-system`: Allow building with `ansi-wl-pprint-1.0.*` and `containers-0.7.*`

RyanGlScott:
This is needed to allow more build plans when building `what4-transition-system` using GHC 9.12.


sauclovian-g (approved):

sauclovian-g:
I did say I had a branch already, but it doesn't matter...

PR #366, opened 2026-04-30
Check for linear arithmetic support in `floatToReal`

### LINKED ISSUE #361: Bitwuzla crashes (`Parse exception: Failed reading: empty`) when given `iFloatToReal` instead of giving a proper error message
Normally, What4 is reasonably good about checking if you are about to give it a query that involves linear arithmetic, and if the underlying solver does not support it, then What4 will raise an exception before it sends the query to the solver. For instance, this can be seen in the following Cryptol interaction that uses a What4-based Bitwuzla solver interaction. Note that Bitwuzla does not support linear arithmetic, and What4 catches this upfront:

```
Float> :set prover=w4-bitwuzla
Float> :prove \(x : Integer) -> x == 0

What4 exception:
user error (Bitwuzla does not support the linear arithmetic term generated at :1:0)
```

This check doesn't always occur, however, and when it doesn't, then What4 will happily send an unsupported query to the solver, causing pandemonium as a result. For example, in the following interaction:

```
Float> :m Float
Loading module Cryptol
Loading module Float
Float> :prove \(x : Float32) -> fpToRational x == 0

What4 exception:
Parse exception: Failed reading: empty
```

This time, What4 simply gives an inscrutable `Parse exception: Failed reading: empty` error. The culprit is ultimately that we are giving a query to Bitwuzla that it does not support. This can be seen in the following setup that uses the What4 API:

```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE TypeOperators #-}
module Main (main) where

import Data.Foldable (forM_)
import System.IO (stdout)

import What4.Expr
         ( Expr, FloatModeRepr(..), newExprBuilder
         , BoolExpr, groundEval, EmptyExprBuilderState(..) )
import What4.Interface
import What4.InterpretedFloatingPoint
import What4.Solver
import What4.Protocol.Online (SolverProcess, checkAndGetModel, startSolverProcess, shutdownSolverProcess, solverConn)
import What4.Protocol.SMTLib2 (Writer, assume, defaultFeatures)

main :: IO ()
main = do
  extendConfig bitwuzlaOptions (getConfiguration sym)
  s <-
    -- startSolverProcess (defaultFeatures Bitwuzla) Nothing sym
    startSolverProcess (defaultFeatures Bitwuzla) (Just stdout) sym

  fp1 <- freshFloatConstant sym (safeSymbol "fp1") SingleFloatRepr
  fp2 <- freshFloatConstant sym (safeSymbol "fp2") SingleFloatRepr
  r1 <- iFloatToReal @_ @SingleFloat sym fp1
  r2 <- iFloatToReal @_ @SingleFloat sym fp2
  {-
  r1 <- freshConstant sym (safeSymbol "r1") BaseRealRepr
  r2 <- freshConstant sym (safeSymbol "r2") BaseRealRepr
  -}
  same <- realEq sym r1 r2

  checkModel s same [("r1", r1)]
  _ <- shutdownSolverProcess s
  pure ()

checkModel ::
  SolverProcess t (Writer Bitwuzla) ->
  BoolExpr t ->
  [(String, Expr t BaseRealType)] ->
  IO ()
checkModel s f es = do
  -- Assume f is true.
  assume (solverConn s) f
  checkAndGetModel s "checkModel" >>= \result ->
    case result of
      Sat ge -> do
        forM_ es $ \(nm, e) -> do
          v <- groundEval ge e
```

If we run this, we get:

```
$ runghc Bug.hs
(set-option :print-success true)
; success
(set-option :produce-models true)
; success
(set-option :global-declarations true)
; success
(set-option :produce-unsat-cores true)
; success
(set-logic ALL)
; success
(push 1)
; success
; :1:0
(declare-fun x!0 () (_ FloatingPoint 8 24))
; success
(define-fun x!1 () Real (fp.to_real x!0))
*** Exception: Parse exception: Failed reading: empty
```

This prints out the offending SMT-LIB query. If we pass this to Bitwuzla, then it fails with:

```
$ bitwuzla test.smt2
success
success
success
success
success
success
success
[error] test.smt2:16:20: invalid sort 'Real'
```

What4 doesn't recognize the `invalid sort 'Real'` error message, which is ultimately what causes the `Parse exception: Failed reading: empty` exception. That being said, it should never get to that point: What4 should see ahead of time that `fp.to_real` returns a `Real` and check that the underlying solver supports linear arithmetic. As we saw earlier, this happens for certain types of queries, but for whatever reason, `fp.to_real` doesn't cause this check to fire. It arguably should.

RyanGlScott:
This ensures that solvers which do not support linear arithmetic (e.g., Bitwuzla) will raise a proper error message when encountering `floatToReal` operations instead of throwing an inscrutable parse exception.

Fixes #361.

### DIFF
--- a/what4/src/What4/Protocol/SMTWriter.hs
+++ b/what4/src/What4/Protocol/SMTWriter.hs
@@ -2677,6 +2677,7 @@ appSMTExpr ae = do
       xe <- mkBaseExpr x
       freshBoundTerm (BVTypeMap w) $ floatToSBV (natValue w) r xe
     FloatToReal x -> do
+      checkLinearSupport i
       xe <- mkBaseExpr x
       freshBoundTerm RealTypeMap $ floatToReal xe
     FloatSpecialFunction{} -> unsupportedTerm i
--- a/what4/CHANGES.md
+++ b/what4/CHANGES.md
@@ -4,6 +4,9 @@
   were lost after @reset-assertions@. These constraints are now properly
   re-asserted after reset. This fix ensures that variables cached with
   @DeleteNever@ maintain their necessary constraints across solver resets.
+* What4 now gives a proper error message when preparing to send a query to
+  Bitwuzla involving the `floatToReal` operation (which Bitwuzla does not
+  support) instead of throwing a parse exception.
 * Replace `lens` with `microlens-{,-mtl,-th}`.
 
 # 1.7.3 -- 2026-01-26


what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -2677,6 +2677,7 @@ appSMTExpr ae = do
       xe <- mkBaseExpr x
       freshBoundTerm (BVTypeMap w) $ floatToSBV (natValue w) r xe
     FloatToReal x -> do
+      checkLinearSupport i
```
langston-barrett, line 2680: "linear" is a bit general of a name here, we should probably distinguish [LIA and LRA](https://smt-lib.org/logics.shtml) more clearly... but that's an existing problem.

langston-barrett (approved):

PR #371, opened 2026-05-14
what4-domains: Split out abstract domain package 

### LINKED ISSUE #370: Split out abstract domain package
I'd like to add a bitwise domain to Macaw, much like What4's. I don't want to depend on all of What4 in macaw-base, which otherwise does not use anything SMT-related. Let's split out a `what4-domains` package so that we can more easily reuse just the abstract domains in other projects.

langston-barrett:
Fixes #370


what4-domains/what4-domains.cabal
```diff
@@ -0,0 +1,185 @@
...
+  exposed-modules:
+    What4.Domains.BV
+    What4.Domains.BV.Arith
+    What4.Domains.BV.Bitwise
+    What4.Domains.BV.XOR
+    What4.Domains.Internal
+    What4.Domains.Verification
+
+  other-modules:
+    What4.Domains.Arithmetic
```
langston-barrett, line 169: This is just duplicated because it's not worth worrying about exporting it.

langston-barrett:
Huh, weird CI failure:
```
Building library for what4-domains-0.1.0.0..

on the commandline: error: [GHC-93741] [-Wunrecognised-warning-flags, Werror=unrecognised-warning-flags]
    unrecognised warning flag: -Werror=inconsistent-flags
```
But this is GHC 9.8.1, which is exactly [when that flag was introduced](https://downloads.haskell.org/ghc/9.10.1-alpha2/docs/users_guide/using-warnings.html?highlight=warnings#ghc-flag--Winconsistent-flags).

https://github.com/GaloisInc/what4/actions/runs/25921251577/job/76190716416?pr=371

RyanGlScott:
I think that part of the GHC User's Guide is just wrong. https://github.com/ghc/ghc/commit/21a906c28da497c2b8390de75270357a7f80e5a7 (the commit which introduced `-Winconsistent-flags`) shows as first appearing in GHC 9.10.1, not 9.8.1.

langston-barrett:
Oh that's interesting - I copy/pasted this block from elsewhere, but I haven't run into this problem with it before...

Thanks for the fact-check 😁

what4-domains/what4-domains.cabal
```diff
@@ -0,0 +1,98 @@
+Cabal-version: 2.4
+Name:          what4-domains
+Version:       0.1.0.0
```
RyanGlScott: I think it would be worth using the development version convention that `what4` uses by giving this a `.99` version component at the end. If we intend for the first Hackage release to be `0.1`, then this means that we should use `0.0.0.0.99` as the version number for the time being.

what4-domains/what4-domains.cabal
```diff
@@ -0,0 +1,98 @@
...
+Author:        Galois Inc.
+Maintainer:    rscott@galois.com, kquick@galois.com, langston@galois.com
+Copyright:     (c) Galois, Inc 2014-2024
```
RyanGlScott: Worth refreshing the copyright year while you're in the neighborhood?

what4-domains/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright (c) 2013-2023 Galois Inc.
```
RyanGlScott: Worth refreshing the copyright year while you're in the neighborhood?

what4-domains/what4-domains.cabal
```diff
@@ -0,0 +1,98 @@
...
+
+Extra-doc-files:
+  doc/bvdomain.cry
+  doc/arithdomain.cry
+  doc/bitsdomain.cry
+  doc/xordomain.cry
+
+source-repository head
+  type: git
+  location: https://github.com/GaloisInc/what4
```
RyanGlScott, line 25: Worth adding a `subdir` here:

```cabal
source-repository head
  ...
  subdir: what4-domains
```

what4-domains/test/BVDomTests.hs
RyanGlScott: Typo: s/bvdoman/bvdomain/

what4-domains/doc/arithdomain.cry
RyanGlScott, line 1: Note that these `.cry` files have an accompanying `what4/doc/README.md` file, which I think should be moved to `what4-domains/doc/README.md` now. Make sure to also update the module names referenced in that `README` to reflect their new names under `what4-domains`.

RyanGlScott (commented):
This is certainly worth doing, although it will require some churn in downstream `cabal.project` files to point to the new directory where `what4-domains` lives in. I'm hoping that you'll help adapt to this churn in some of our downstream projects :)

langston-barrett:
Yep, I'll be happy to!

RyanGlScott (approved):

PR #375, opened 2026-05-15
domains: Strided interval domain

### LINKED ISSUE #374: feat(domains): Strided interval domain
Macaw has one of these, let's port it into `what4-domains` to make it more discoverable and reusable. We could use this as part of a reduced product domain (#372).

Prior art:

- https://github.com/statinf-otawa/otawa-clp
- https://github.com/ucsb-seclab/sasi
- https://github.com/trailofbits/maat/blob/ac068c6cc7120cdab817ba97d362203dccb9e29b/src/include/maat/expression.hpp#L125
- https://github.com/angr/claripy/blob/3a6ac358ab319c997b1b7c96929f52dff99c1f0a/claripy/backends/backend_vsa/strided_interval.py#L152

langston-barrett:
Fixes #374. Based on the implementation in Macaw, but with a few more operations.

Major differences from Macaw's strided interval:

- Handles wrapping
- Stores a mask, not a width

Needs:

- [ ] Table comparing operations to those in Macaw
- [ ] Bitwise operations via bitwise domain

### DIFF
--- a/what4-domains/doc/arithdomain.cry
+++ b/what4-domains/doc/arithdomain.cry
@@ -11,6 +11,8 @@ bit width) by loading this file in cryptol and entering ":prove".
 */
 module arithdomain where
 
+import bitsdomain as B
+
 ////////////////////////////////////////////////////////////
 // Library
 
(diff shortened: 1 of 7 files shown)


langston-barrett:
This is currently heavily based on the one in Macaw, but I think it's worth improving. I'd like to:

- [x] Support all of the same operations as `Arith.hs` (e.g., signed bounds/extension)
- [x] Support wrapping intervals like `Arith.hs` does
- [x] Add differential tests against `Arith.hs` for `stride = 1` intervals

langston-barrett:
This is blocked on #380 for differential tests with `Arith.hs`.

langston-barrett:
I have an idea for a better domain.

PR #379, opened 2026-05-18
Fix `floatToInteger`'s behavior when using `RNA`

### LINKED ISSUE #377: `groundEval` and SMT solvers disagree on the output of rounding floats using `RNA`
In the following `what4` program:

```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GADTs #-}
module Main where

import Data.Foldable (forM_)
import LibBF (bfFromDouble)

import What4.Expr
import What4.Interface
import What4.Solver
import What4.Protocol.SMTLib2
         (assume, sessionWriter, runCheckSat)


z3executable :: FilePath
z3executable = "z3"

main :: IO ()
main = do
  extendConfig z3Options (getConfiguration sym)
  let w :: NatRepr 32
      w = knownNat
  let fpp :: FloatPrecisionRepr Prec32
      fpp = knownRepr
  lit <- floatLit sym fpp (bfFromDouble 0.25)
  f <- freshConstant sym (safeSymbol "f") (BaseFloatRepr fpp)
  p1 <- floatEq sym lit f
  bv1 <- floatToBV sym w RNA lit
  bv2 <- floatToBV sym w RNA f
  bv3 <- freshConstant sym (safeSymbol "bv") (BaseBVRepr w)
  p2 <- bvEq sym bv2 bv3
  p <- andPred sym p1 p2
  checkModel sym p [("bv1", bv1), ("bv2", bv2), ("bv3", bv3)]

checkModel ::
  ExprBuilder t st fs ->
  BoolExpr t ->
  [(String, Expr t (BaseBVType 32))] ->
  IO ()
checkModel sym f es = do
  withZ3 sym z3executable defaultLogData $ \session -> do
    assume (sessionWriter session) f
    runCheckSat session $ \result ->
      case result of
        Sat (ge, _) -> do
          forM_ es $ \(nm, e) -> do
            v <- groundEval ge e
```

We do the following:

1. We create a fresh single-precision float `f`. We also create a single-precision float literal `lit` with the value `0.25`. We assert that `f` equals `lit`.
2. We convert both `lit` and `f` to unsigned bitvectors using the `RNA` (round to nearest integer, ties away from zero) rounding mode, calling the results `bv1`, and `bv2`, respectively.
3. Finally, we create a fresh bitvector `bv3` and assert that it is equal to `bv2`.

I would expect this query to be satisfiable, and `what4` should report a model where all of `bv1`, `bv2`, and `bv3` have the value `0`. This is because the nearest integer to `0.25` is `0`. Despite this, `what4` reports a model where these values are _different_:

```
$ runghc Main.hs
Satisfiable, with model:
  bv1 := BV 1
  bv2 := BV 0
  bv3 := BV 0
```

Note that `bv1` incorrectly reports its value as `1`, not `0`. This is because of the following bit of erroneous code used behind the scenes when grounding `FloatToBV` operations:

https://github.com/GaloisInc/what4/blob/a1549896ceb01765a8c1eee097ce70ef43a5c38d/what4/src/What4/Utils/FloatHelpers.hs#L92-L101

The code in the `RNA` case is suspect, as it incorrectly implements the `RNA` rounding mode for `Rational` values. I think that we should use the `roundAway` implementation here in `what4` instead:

https://github.com/GaloisInc/what4/blob/eb045510fda45f77cb1ed5950bc51495597b36a6/what4/src/What4/Utils/Arithmetic.hs#L128-L130

This is essentially the same bug as in https://github.com/GaloisInc/cryptol/issues/2044, which isn't surprising---I suspect that the code in `cryptol` was adapted from `what4` (or vice versa).

RyanGlScott:
This ports the fix from https://github.com/GaloisInc/cryptol/pull/2047 over to `what4`.

Fixes #377.

### DIFF
--- a/what4/CHANGES.md
+++ b/what4/CHANGES.md
@@ -17,6 +17,8 @@
   Bitwuzla involving the `floatToReal` operation (which Bitwuzla does not
   support) instead of throwing a parse exception.
 * Replace `lens` with `microlens-{,-mtl,-th}`.
+* Fix a bug where `what4` could produce incorrect models involving the
+  `floatToBV` operation when using the `RNA` rounding mode.
 
 # 1.7.3 -- 2026-01-26
 
--- a/what4/src/What4/Utils/FloatHelpers.hs
+++ b/what4/src/What4/Utils/FloatHelpers.hs
@@ -14,6 +14,7 @@ import LibBF
 
 import What4.BaseTypes
 import What4.Panic (panic)
+import What4.Utils.Arithmetic (roundAway)
 
 -- | Rounding modes for IEEE-754 floating point operations.
 data RoundingMode
@@ -95,7 +96,7 @@ floatToInteger r fp =
   do rat <- floatToRational fp
      pure case r of
             RNE -> round rat
-            RNA -> if rat > 0 then ceiling rat else floor rat
+            RNA -> roundAway rat
             RTP -> ceiling rat
             RTN -> floor rat
             RTZ -> truncate rat
--- a/what4/test/ExprBuilderSMTLib2.hs
+++ b/what4/test/ExprBuilderSMTLib2.hs
@@ -1161,6 +1161,48 @@ issue329Test sym solver = do
 
       _ -> fail "expected satisfible model"
 
+-- | A regression test for #377.
+issue377Test ::
+  OnlineSolver solver =>
+  SimpleExprBuilder t fs ->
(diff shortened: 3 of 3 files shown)


langston-barrett (approved):
I don't know a ton about floats generally or especially about floats in SMT-Lib, but looks reasonable!

PR #380, opened 2026-05-18
fix(domains): Fix srem returning improper value when divisor is {0}

langston-barrett:
When the divisor was the singleton {0}, the non-singleton-quotient branch built `BVDInterval mask _ (rh - rl)` with rh < rl, producing a negative size. Short-circuit on b = {0} and widen to BVDAny.


langston-barrett:
Oof, I've realized that there's a deeper issue lurking... our abstract domains do not relfect SMT-Lib semantics for division by zero. Will post an issue about this.

langston-barrett:
This is blocked on #382.

RyanGlScott (approved):

PR #383, opened 2026-05-18
feat: SMT-Lib bitvector division

### LINKED ISSUE #382: feat: SMT-Lib-compatible division operators
`What4.Domains.BV.Arith.{udiv, urem, sdiv, srem}` and the symbolic `bvUdiv`/`bvUrem`/`bvSdiv`/`bvSrem` simplifications that consume their abstract-value results produce answers that are inconsistent with the SMT-LIB spec when the divisor is the singleton domain `{0}`. This can be seen by running `w4smt2` and Z3 on a few SMT-Lib inputs.

Tested at `origin/master` (`4e1355a2`).

## SMT-LIB specification

From the [`FixedSizeBitVectors` theory](https://smt-lib.org/theories-FixedSizeBitVectors.shtml) (version 2.7):

> ```
> [[(bvudiv s t)]] := if bv2nat([[t]]) = 0
>                     then λx:[0, m). 1
>                     else nat2bv[m](bv2nat([[s]]) div bv2nat([[t]]))
>
> [[(bvurem s t)]] := if bv2nat([[t]]) = 0
>                     then [[s]]
>                     else nat2bv[m](bv2nat([[s]]) mod bv2nat([[t]]))
> ```

i.e.

- `(bvudiv s 0)` is the all-ones bitvector of width `m`.
- `(bvurem s 0)` is `s` itself.

The theory definition does not cover `bvsdiv`/`bvsrem`; those are part of the `QF_BV` logic extension and are implemented by every mainstream solver (Z3, CVC5, Bitwuzla, Yices) consistently as:

- `(bvsdiv s 0)` is `(if (bvslt s #x0) #x1 (bvneg #x1))` — i.e. all-ones when the dividend is non-negative, `1` when negative.
- `(bvsrem s 0)` is `s`.

## Reproduction

Each of the four `.smt2` files below is a one-line assertion of the form `(= (bvOP #x5 #x0) #x5)`.

```sh
cat > bvudiv.smt2 <<'EOF'
(set-logic QF_BV)
(assert (= (bvudiv #x5 #x0) #x5))
(check-sat)
EOF

cat > bvurem.smt2 <<'EOF'
(set-logic QF_BV)
(assert (= (bvurem #x5 #x0) #x5))
(check-sat)
EOF

cat > bvsdiv.smt2 <<'EOF'
(set-logic QF_BV)
(assert (= (bvsdiv #x5 #x0) #x5))
(check-sat)
EOF

cat > bvsrem.smt2 <<'EOF'
(set-logic QF_BV)
(assert (= (bvsrem #x5 #x0) #x5))
(check-sat)
EOF

for f in bvudiv.smt2 bvurem.smt2 bvsdiv.smt2 bvsrem.smt2; do
    printf "%-15s z3=%-7s w4smt2=%s\n" "$f" \
      "$(z3 -smt2 "$f")" \
      "$(cabal run -v0 exe:w4smt2 -- "$f")"
  done
```
Output:
```
bvudiv.smt2     z3=unsat   w4smt2=sat
bvurem.smt2     z3=sat     w4smt2=unsat
bvsdiv.smt2     z3=unsat   w4smt2=sat
bvsrem.smt2     z3=sat     w4smt2=sat
```

Summarized:

| File           | SMT-LIB value of LHS    | Equal to `#x5`? | Z3      | `w4smt2` (no solver) | Verdict       |
|----------------|-------------------------|-----------------|---------|----------------------|---------------|
| `bvudiv.smt2`  | `#xF` (all-ones)        | no              | `unsat` | `sat`                | **wrong**     |
| `bvurem.smt2`  | `#x5` (the dividend)    | yes             | `sat`   | `unsat`              | **wrong**     |
| `bvsdiv.smt2`  | `#xF` (all-ones, s ≥ 0) | no              | `unsat` | `sat`                | **wrong**     |
| `bvsrem.smt2`  | `#x5` (the dividend)    | yes             | `sat`   | `sat`                | accidentally right |

Three of four cases reach a literal answer that contradicts the SMT-LIB spec. The fourth (`bvsrem`) happens to land on the right answer by accident, but the abstract domain it produces internally is unrelated to the SMT-LIB-mandated value.

## Diagnosis

In `what4-domains/src/What4/Domains/BV/Arith.hs`, all four ops compute the result domain from the divisor's bounds using `max 1 _` to side-step division by zero:

```haskell
udiv a b = interval mask ql (qh - ql)
  where
    ...
    ql = al `div` max 1 bh   -- "assume that division by 0 does not happen"
    qh = ah `div` max 1 bl
```

That assumption is inconsitent with SMT-LIB: SMT-LIB defines a value for division by zero, and the abstract domain should include it. With `b = {0}`, `udiv {5} {0}` computes `ql = 5 / max 1 0 = 5`, `qh = 5`, and returns the singleton `{5}` - the *dividend*, not all-ones. The symbolic-eval machinery in `What4.Expr.AppTheory` then sees an `asSingleton` result and constant-folds the expression to `#x5`.

## Suggestion

This isn't a bug. `bvUdiv`/`bvUrem`/`bvSdiv`/`bvSrem` all document that they are undefined when dividing by zero. Let's keep them undefined at zero so that we can keep the tight abstract values we produce now. However, let's offer SMT-Lib compatible abstract transformers and `ExprBuilder` methods for cases where we explicitly want to match SMT-Lib. Let's use those in `w4smt2`.

langston-barrett:
Fixes #382.


RyanGlScott (approved):
Nice! I like that this approach doesn't add more constructors to `App`, which should minimize downstream churn.

Do make sure to mention the new `IsExprBuilder` methods in the changelog, but otherwise LGTM.

PR #384, opened 2026-05-19
`What4.SFloat`: Add missing API operations

### LINKED ISSUE #376: `What4.SFloat`: Add missing API operations
The [`What4.SFloat`](https://github.com/GaloisInc/what4/blob/a1549896ceb01765a8c1eee097ce70ef43a5c38d/what4/src/What4/SFloat.hs) module defines operations over floating-point values where the exponent and significand sizes are only known at runtime. This is primarily useful for downstream tools like Cryptol, which pass around float values internally in a "dynamically typed" fashion.

In general, the operations in `What4.SFloat` are in correspondence to the floating-point operations provided in `What4.Interface`, but there are some notable omissions. In particular, `What4.SFloat` lacks counterparts to the following `What4.Interface` operations:

* `floatNZero`
* `floatPZero`
* `floatRem`
* `floatNe`
* `floatFpApart`
* `floatFpUnordered`
* `floatLe`
* `floatGe`
* `floatIsPos`
* `floatCast`
* `bvToFloat` (will likely require `What4.SFloat` to define)
* `sbvToFloat` (will likely require `What4.SFloat` to define)
* `floatToBV` (will likely require `What4.SFloat` to define)
* `floatToSBV` (will likely require `What4.SFloat` to define)

RyanGlScott:
This achieves more complete coverage of floating-point operations compared to what is available in `What4.Interface`.

Fixes #376.

### DIFF
--- a/what4/src/What4/SFloat.hs
+++ b/what4/src/What4/SFloat.hs
@@ -18,6 +18,8 @@ module What4.SFloat
     -- * Constants
   , fpFresh
   , fpNaN
+  , fpPosZero
+  , fpNegZero
   , fpPosInf
   , fpNegInf
   , fpFromLit
@@ -30,9 +32,14 @@ module What4.SFloat
     -- * Relations
   , SFloatRel
   , fpEq
+  , fpNe
   , fpEqIEEE
+  , fpNeIEEE
+  , fpLeIEEE
   , fpLtIEEE
+  , fpGeIEEE
   , fpGtIEEE
+  , fpUnordered
 
     -- * Arithmetic
   , SFloatBinArith
@@ -43,6 +50,7 @@ module What4.SFloat
   , fpSub
   , fpMul
   , fpDiv
+  , fpRem
   , fpMin
   , fpMax
   , fpFMA
@@ -54,12 +62,18 @@ module What4.SFloat
   , fpFromRational
   , fpToRational
   , fpFromInteger
+  , fpCast
+  , fpFromBV
+  , fpFromSBV
+  , fpToBV
+  , fpToSBV
 
     -- * Queries
   , fpIsInf
   , fpIsNaN
   , fpIsZero
   , fpIsNeg
+  , fpIsPos
   , fpIsSubnorm
   , fpIsNorm
 
@@ -70,6 +84,7 @@ module What4.SFloat
 
 import Control.Exception
 import LibBF (BigFloat)
+import Numeric.Natural
 
 import Data.Parameterized.Some
 import Data.Parameterized.NatRepr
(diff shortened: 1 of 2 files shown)


what4/src/What4/SFloat.hs
```diff
@@ -179,6 +194,27 @@ fpNaN sym e p
   | Just (Some fpp) <- fpRepr e p = SFloat <$> floatNaN sym fpp
   | otherwise = unsupported "fpNaN" e p
 
+-- | Positive zero
+fpPosZero ::
```
langston-barrett: Seems like these mostly delegate to methods from `IsExprBuilder`, would it be worth referring readers to those docs?
```suggestion
-- | Positive zero (see 'floatPZero')
fpPosZero ::
```

what4/src/What4/SFloat.hs
```diff
@@ -433,7 +492,59 @@ fpToRational sym fp =
...
+-- bitvector) is non-zero.
+fpToSBV ::
+  IsExprBuilder sym =>
+  sym -> Natural -> RoundingMode -> SFloat sym -> IO (SWord sym)
+fpToSBV sym n r (SFloat x) =
+  case mkNatRepr n of
+    Some nr ->
+      case isPosNat nr of
+        Nothing -> panic "fpToSBV" ["bit width must be non-zero"]
+        Just LeqProof -> DBV <$> floatToSBV sym nr r x
```
langston-barrett: Perhaps we should consider versions of these that return `Maybe`, along with `panic`ing wrappers? Seems more flexible for consumers if they can choose. Alternatively/also, we could consider versions that take `1 <= w => NatRepr w`?

langston-barrett (approved):

what4/src/What4/SFloat.hs
RyanGlScott: > Perhaps we should consider versions of these that return `Maybe`, along with panicing wrappers?

The functions in this module are already very partial and will readily raise errors if you fail to uphold various preconditions (e.g., if you pass invalid exponent/significand sizes). While we could duplicate every function to offer a version that returns `Maybe`, I don't think it would offer much value. The only downstream consumers of this functionality are Cryptol and SAW, and they will ultimately propagate the errors that these functions raise anyway.

> Alternatively/also, we could consider versions that take `1 <= w => NatRepr w`?

Similarly, I don't think this would offer much value for Cryptol or SAW, at least in the way they're currently designed. At present, their internals are quite dynamic, and there would be placed where you'd need to erase the `w` in `NatRepr w` anyway. At that point, we might as well use `Natural`.

what4/src/What4/SFloat.hs
RyanGlScott: In https://github.com/GaloisInc/what4/pull/384/commits/b9356be3687eab684b1a2255dd037eb30137af2d, I've made an effort to document the relevant `IsExprBuilder` methods in places where one function obviously corresponds to a single method.

PR #390, opened 2026-05-20
feat(domains): Precise division, fast multipliction for bitwise domain

langston-barrett:
Tnum multiplication is O(w^2), add a fast bounds-based O(w) verison. Conversely, tnum division as-implemented was a fast bounds-based O(w) version, add the precise "schoolbook" O(w^2) algorithm, just like for multiplication.

### DIFF
--- a/what4-domains/doc/bitsdomain.cry
+++ b/what4-domains/doc/bitsdomain.cry
@@ -170,18 +170,77 @@ bneg a = add (bnot a) (singleton 1)
 sub : {n} (fin n, n >= 1) => Dom n -> Dom n -> Dom n
 sub a b = add a (bneg b)
 
+// scale uses the precise (shift-and-add) multiplication so that
+// constant-by-domain products preserve bit-level structure.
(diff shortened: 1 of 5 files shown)


what4-domains/doc/bitsdomain.cry
```diff
@@ -466,6 +662,9 @@ property a10 = correct_udiv`{8}
 property a11 = correct_urem`{8}
 property a12 = correct_sdiv`{8}
 property a13 = correct_srem`{8}
+property a14 = correct_mulPrecise`{8}  // NB: takes 60s
```
RyanGlScott, line 665: Takes 60 seconds to `:check`? To `:prove`?

RyanGlScott (approved):
I'll admit that this is pretty far over my head, so I can't offer a very detailed review. I do appreciate the extensive testing that this adds.

what4-domains/doc/bitsdomain.cry
langston-barrett: `:prove`!

PR #392, opened 2026-05-20
perf: Use helpers from `base`, `ghc-bignum` for arithmetic

### LINKED ISSUE #389: perf: Replace `Utils.Arithmetic` helpers with builtins from `base`, `ghc-bignum`
`What4.Utils.Arithmetic` defines a bunch of helper functions:

https://github.com/GaloisInc/what4/blob/e4d678718585d1f26e6d8207d69f3fce8a9ee4a0/what4/src/What4/Utils/Arithmetic.hs

But many like `lg` have [equivalents in `ghc-bignum`](https://hackage-content.haskell.org/package/ghc-bignum-1.3/docs/GHC-Num-Integer.html#v:integerLog2) or ones like `ctz` have [equivalents in `base`](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Data-Bits.html#v:countTrailingZeros). These are almost surely faster than our hand-rolled versions. We should audit this module and replace these with the built-in versions. `ghc-bignum` will require some CPP (it is only available with GHC 9.0+), but `Arithmetic.hs` is a small module, so it seems worth it.

langston-barrett:
Fixes #389.


what4/src/What4/Utils/Arithmetic.hs
```diff
@@ -31,40 +36,64 @@ import Data.Ratio
...
--- | Returns ceil of log base 2.
+-- | @intLog2 n@ for @n >= 1@: floor of base-2 logarithm. Undefined for
+-- @n <= 0@. On GHC 9.0+ this delegates to a fast primop in @ghc-bignum@;
+-- on earlier GHCs it falls back to 'lg'.
+intLog2 :: Integer -> Int
+intLog2 = intLog2Opt
+{-# INLINE intLog2 #-}
+
+-- | Returns ceil of log base 2. Polymorphic over bit-like types.
 --   We define @lgCeil 0 = 0@
```
RyanGlScott: And `lgCeil 1 = 0`, judging from the implementation?

what4-domains/src/What4/Domains/Arithmetic/Internal.hs
```diff
@@ -0,0 +1,105 @@
...
+   | otherwise = i
+{-# INLINABLE ctzRef #-}
+
+-- | Reference implementation: Count leading zeros using bit testing loop
+clzRef :: NatRepr w -> Integer -> Integer
+clzRef w x = go 0
+ where
+ go !i
+   | i < toInteger (natValue w) && testBit x (widthVal w - fromInteger i - 1) == False = go (i + 1)
+   | otherwise = i
```
RyanGlScott: I think the bang patterns on the `i` arguments are unnecessary, as the first guard evaluates `i` to WHNF anyway when it performs the `i < toInteger (natValue w)` comparison.

RyanGlScott (approved):

what4-domains/src/What4/Domains/Arithmetic/Internal.hs
langston-barrett: I pushed a follow-up that addresses these and also adds some more helpers that reduce the direct bit-manipulation inline.

RyanGlScott (approved):

PR #398, opened 2026-05-21
feat(domains): General shifts for the bitwise domain

### LINKED ISSUE #395: feat(domains): Shifts by non-constant amounts for the bitwise domain
We should do whatever LLVM's `KnownBits` does: https://github.com/llvm/llvm-project/blob/f46af8471d7934614a7d8efbb05deb9a657a75f8/llvm/lib/Support/KnownBits.cpp

langston-barrett:
Fixes #395.

### DIFF
--- a/what4-domains/src/What4/Domains/BV/Bitwise.hs
+++ b/what4-domains/src/What4/Domains/BV/Bitwise.hs
@@ -7,6 +7,7 @@ Maintainer  : huffman@galois.com
 Provides a bitwise implementation of bitvector abstract domains.
 -}
 
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE KindSignatures #-}
@@ -55,6 +56,16 @@ module What4.Domains.BV.Bitwise
   , ashr
   , rol
   , ror
+  , shlAbstract
+  , lshrAbstract
+  , ashrAbstract
+  , rolAbstract
+  , rorAbstract
+  , shlAbstractSpec
+  , lshrAbstractSpec
+  , ashrAbstractSpec
+  , rolAbstractSpec
+  , rorAbstractSpec
   -- ** arithmetic
   , add
   , sub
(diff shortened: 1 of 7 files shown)


what4-domains/src/What4/Domains/BV/Bitwise.hs
```diff
@@ -447,6 +468,269 @@ ashr w (BVBitInterval mask lo hi) y = BVBitInterval mask (shr lo) (shr hi)
   y' = fromInteger (min y (intValue w))
   shr x = ((toSigned w x) `shiftR` y') .&. mask
 
+-- | Conflict ("empty") domain: invariant @lo [= hi@ is violated.
```
RyanGlScott, line 471: I'm not sure how to read `lo [= hi`. What does `[=` indicate?

what4-domains/src/What4/Domains/BV/Bitwise.hs
```diff
@@ -447,6 +468,269 @@ ashr w (BVBitInterval mask lo hi) y = BVBitInterval mask (shr lo) (shr hi)
...
+        union acc (op (fromInteger wI))
+    | otherwise = acc
+
+  collapse d
+    | isConflict d = BVBitInterval mask 0 0
+    | otherwise    = d
+
+-- | /O((bh - bl + 1) · w / W)/ where @W@ is the machine word size.
+-- Shift left by an amount drawn from the domain @b@. See 'foldShifts'
+-- for the algorithm.
```
RyanGlScott: What are `bh`, `bl`, and `w`? (Similar questions for the functions below.)

RyanGlScott (approved):

what4-domains/src/What4/Domains/BV/Bitwise.hs
langston-barrett: That's from the Haddock on `Domain`:
```haskell
-- | A bitwise interval domain, defined via a
--   bitwise upper and lower bound.  The ordering
--   used here to construct the interval is the pointwise
--   ordering on bits.  In particular @x [= y iff x .|. y == y@,
--   and a value @x@ is in the set defined by the pair @(lo,hi)@
--   just when @lo [= x && x [= hi@.
```

what4-domains/src/What4/Domains/BV/Bitwise.hs
langston-barrett: Clarified!

PR #399, opened 2026-05-21
`fpToRational`: Ensure that the denominator is non-zero

RyanGlScott:
Towards an eventual fix for https://github.com/GaloisInc/cryptol/issues/2021 on the Cryptol side (which is where a test case will eventually go).


langston-barrett (approved):
I'm a little surprised the solver would ever pick 0 for the denominator given the equality assertion - is this another instance of SMT-Lib defining division by zero, yielding `0 / 0 = 0`?

RyanGlScott:
Yes, SMT-LIB's `/` function is total on real numbers, which means that `x / 0` is well defined, but it returns an unspecified value. (It could be `0`, but it could also be something else entirely.)

PR #404, opened 2026-05-21
feat(domains): Port some precision improvements from LLVM's `KnownBits`

langston-barrett:


RyanGlScott (approved):
I'm quite happy with how the precision-tracking tests shows concrete improvements at each step. Thanks for adding those!

PR #407, opened 2026-05-22
feat(domains): Additional operations for interval domain via bitwise

### LINKED ISSUE #405: feat(domains): Additional operations for interval domains via bitwise
The `Arith.hs` domain should support bitwise operations and rotations via conversion to `Bitwise.hs` and back, for consumers that predominantly want to use intervals but still want *some* level of support for bitwise ops.

langston-barrett:
Fixes #405.


RyanGlScott (approved):

langston-barrett:
Closing for now, I think the correct answer is a reduced product domain I'm working on.

PR #410, opened 2026-06-03
test(domains): Precision regression tests

langston-barrett:
See the Haddock. This is a good way to test things like #404.


what4-domains/test/PrecisionRegression.hs
```diff
@@ -0,0 +1,29 @@
...
+
+import           Test.Tasty (defaultMain, testGroup)
+
+import           PrecisionRegression.Common (domainTests)
+import qualified PrecisionRegression.Arith as ArithReg
+import qualified PrecisionRegression.Bitwise as BitwiseReg
+
+main :: IO ()
+main = do
+  update <- (== Just "1") <$> lookupEnv "WHAT4_UPDATE_TEST_EXPECTATIONS"
```
RyanGlScott, line 26: Perhaps it would be worth using the `tasty-golden` library here? Test suite executables that use `tasty-golden` come with an `--accept` flag that, when passed, automatically regenerate golden files whose results have changed.

RyanGlScott (approved):

what4-domains/test/PrecisionRegression.hs
langston-barrett: Good call, will do.

langston-barrett: Oh actually, that would lose parallelism - the current implementation uses Tasty's multi-threading to compute the ratio for each operation separately. This is nice because these are fairly expensive computations. A golden test would be one test per domain, losing out on a lot of opportunity here. Or we could have one file per op, but they would each be one line... I'll just keep it as-is for now.

PR #424, opened 2026-07-15
feat(what4): Quasiquoter for `SymExpr`

### LINKED ISSUE #422: A quasiquoter for `SymExpr`
Consider `testRotate1`:

https://github.com/GaloisInc/what4/blob/e1527bc29774c8579a0d93587dc9dc986859dd53/what4/test/ExprBuilderSMTLib2.hs#L515-L532

It's hard to see at a glance the general shape of the expressions being built here, and thus hard to see what the test is testing. The same problem applies to production code.

Template Haskell quasiquoters provide a simple way to transform expressions written in a DSL into Haskell expressions (or types or patterns) at compile-time. We may want to consider developing a DSL and quasiquoter for What4 expressions.

SMT-Lib would be a natural fit, but its language has some warts. For example, bitvector literals are written `(_ bvNNN W)` where `N` is a numeral representing the bitvector's value, and `W` is a width. Since the point of developing this quasiquoter would be to make code more readable, adopting SMT-Lib wholesale might defeat the point.

Of course, `What4.Syntax` syntax would be another option, but it inherits much of the strangeness of SMT-Lib `(_ extract i j)` and is even more verbose (`intadd` instead of `+`). I also find it excessively verbose to type `intle` or `bvadd` when the "int" or "bv" could be inferred. I instead propose we have generic `add`/`le` and use a typeclass to instantiate them appropriately. Of course, we could choose consistency over readability and just adopt this syntax wholesale.

In any case, a quasiquoter would want affordances such as metavariables, which means we're building our own parser. We should probably keep it simple and use S-expressions. If we like it, we could add a more complete, less parenthesized language later.

The exact choice of spelling for each operation is certainly a quite subjective choice. We can expect readers and authors to be familiar with Haskell. We *might* expect them to be familiar with SMT-Lib. We could also take inspiration from C and Crucible, as DSLs for working with bitvectors. Here's a proposal for some spellings that mix in a bit of each of those:

## Boolean

| Operation      | C      | Haskell  | SMT-LIB    | crucible-syntax | What4.Serialize | w4qq      |
|----------------|--------|----------|------------|-----------------|-----------------|-----------|
| and            | `&&`   | `&&`     | `and`      | `and`           | `andp`          | `and`     |
| or             | `\|\|` | `\|\|`   | `or`       | `or`            | `orp`           | `or`      |
| not            | `!`    | `not`    | `not`      | `not`           | `notp`          | `not`     |
| implies        | -      | -        | `=>`       | -               | -               | `implies` |
| xor            | -      | `/=`     | `xor`      | `xor`           | `xorp`          | `xor`     |
| if-then-else   | `?:`   | `if …`   | `ite`      | `if`            | `ite`           | `ite`     |
| equal          | `==`   | `==`     | `=`        | `equal?`        | `=`             | `eq`      |
| not equal      | `!=`   | `/=`     | `distinct` | -               | -               | `ne`      |

`eq`/`ne` and `ite` work on any sort; the operand/branch sorts must match (a GHC
type error otherwise).

## Integer and real

| Operation        | C       | Haskell  | SMT-LIB | crucible-syntax | What4.Serialize | w4qq  |
|------------------|---------|----------|---------|-----------------|-----------------|-------|
| add              | `+`     | `+`      | `+`     | `+`             | `intadd`        | `add` |
| subtract         | `-`     | `-`      | `-`     | `-`             | -               | `sub` |
| negate (unary)   | `-`     | `negate` | `-`     | `negate`        | -               | `neg` |
| multiply         | `*`     | `*`      | `*`     | `*`             | `intmul`        | `mul` |
| divide (int)     | `/`     | `div`    | `div`   | `/`             | `intdiv`        | `div` |
| modulus (int)    | `%`     | `mod`    | `mod`   | `mod`           | `intmod`        | `mod` |
| divide (real)    | `/`     | `/`      | `/`     | `/`             | -               | `div` |
| absolute value   | `abs()` | `abs`    | `abs`   | `abs`           | `intabs`        | `abs` |
| less than        | `<`     | `<`      | `<`     | `<`             | -               | `lt`  |
| less or equal    | `<=`    | `<=`     | `<=`    | `<=`            | `intle`         | `le`  |
| greater than     | `>`     | `>`      | `>`     | -               | -               | `gt`  |
| greater or equal | `>=`    | `>=`     | `>=`    | -               | -               | `ge`  |
| exponent         | -       | `^`/`**` | -       | -               | -               | - (no builder) |

`div` is overloaded (Int → `intDiv`, Real → `realDiv`, BV → see `udiv`/`sdiv`);
`mod` is Integer only. Int and real are never mixed implicitly (no coercion).
`What4.Serialize` has no real-arithmetic operators.

## Bitvector - arithmetic

| Operation           | C   | Haskell  | SMT-LIB  | crucible-syntax | What4.Serialize | w4qq   |
|---------------------|-----|----------|----------|-----------------|-----------------|--------|
| add                 | `+` | `+`      | `bvadd`  | `+`             | `bvadd`         | `add`  |
| subtract            | `-` | `-`      | `bvsub`  | `-`             | `bvsub`         | `sub`  |
| negate (unary)      | `-` | `negate` | `bvneg`  | `negate`        | `bvneg`         | `neg`  |
| multiply            | `*` | `*`      | `bvmul`  | `*`             | `bvmul`         | `mul`  |
| divide, unsigned    | `/` | `div`    | `bvudiv` | `/`             | `bvudiv`        | `udiv` |
| remainder, unsigned | `%` | `mod`    | `bvurem` | `mod`           | `bvurem`        | `urem` |
| divide, signed      | `/` | `div`    | `bvsdiv` | `/$`            | `bvsdiv`        | `sdiv` |
| remainder, signed   | `%` | `mod`    | `bvsrem` | `smod`          | `bvsrem`        | `srem` |

`add`/`sub`/`neg`/`mul` are the same overloaded names as Integer/Real (two's
complement makes them sign-agnostic). What4 has no unsigned-`mod` distinct from
`urem`, and signed `bvsmod` is unimplemented - hence `urem`/`srem`, not
`umod`/`smod`.

## Bitvector - comparison

| Operation             | C    | Haskell | SMT-LIB | crucible-syntax | What4.Serialize | w4qq  |
|-----------------------|------|---------|---------|-----------------|-----------------|-------|
| less than, unsigned   | `<`  | `<`     | `bvult` | `<`             | `bvult`         | `ult` |
| less/equal, unsigned  | `<=` | `<=`    | `bvule` | `<=`            | `bvule`         | `ule` |
| greater, unsigned     | `>`  | `>`     | `bvugt` | -               | `bvugt`         | `ugt` |
| greater/eq, unsigned  | `>=` | `>=`    | `bvuge` | -               | `bvuge`         | `uge` |
| less than, signed     | `<`  | `<`     | `bvslt` | `<$`            | `bvslt`         | `slt` |
| less/equal, signed    | `<=` | `<=`    | `bvsle` | `<=$`           | `bvsle`         | `sle` |
| greater, signed       | `>`  | `>`     | `bvsgt` | -               | `bvsgt`         | `sgt` |
| greater/eq, signed    | `>=` | `>=`    | `bvsge` | -               | `bvsge`         | `sge` |
| equal                 | `==` | `==`    | `=`     | `equal?`        | `bveq`          | `eq`  |
| not equal             | `!=` | `/=`    | `distinct`| -             | `bvne`          | `ne`  |

(C and Haskell distinguish signed/unsigned by the operand type, e.g. `int8_t`
vs `uint8_t`, `Int8` vs `Word8` - not by the operator.)

## Bitvector - bitwise

| Operation  | C   | Haskell        | SMT-LIB | crucible-syntax | What4.Serialize | w4qq  |
|------------|-----|----------------|---------|-----------------|-----------------|-------|
| and        | `&` | `.&.`          | `bvand` | `bv-and`        | `bvand`         | `and` |
| or         | `\|`| `.\|.`         | `bvor`  | `bv-or`         | `bvor`          | `or`  |
| xor        | `^` | `xor` / `.^.`  | `bvxor` | `bv-xor`        | `bvxor`         | `xor` |
| complement | `~` | `complement`   | `bvnot` | `bv-not`        | `bvnot`         | `not` |

`and`/`or`/`xor`/`not` are the same overloaded names as the boolean operators;
on a bitvector they are bitwise, on a `Bool` they are logical.

## Bitvector - shift, rotate

| Operation                | C    | Haskell        | SMT-LIB        | crucible-syntax | What4.Serialize | w4qq   |
|--------------------------|------|----------------|----------------|-----------------|-----------------|--------|
| shift left               | `<<` | `shiftL`       | `bvshl`        | `shl`           | `bvshl`         | `shl`  |
| shift right, logical     | `>>` | `shiftR` (Word)| `bvlshr`       | `lshr`          | `bvlshr`        | `lshr` |
| shift right, arithmetic  | `>>` | `shiftR` (Int) | `bvashr`       | `ashr`          | `bvashr`        | `ashr` |
| rotate left              | -    | `rotateL`      | `rotate_left`  | -               | -               | `rol`  |
| rotate right             | -    | `rotateR`      | `rotate_right` | -               | -               | `ror`  |

## Bitvector - literals and structural

| Operation      | C   | Haskell         | SMT-LIB               | crucible-syntax       | What4.Serialize       | w4qq                        |
|----------------|-----|-----------------|-----------------------|-----------------------|-----------------------|-----------------------------|
| binary literal | -   | `0b0101`        | `#b0101`              | -                     | `#b0101`              | `#b0101`                    |
| hex literal    | `0x`| `0xFF`          | `#xFF`                | -                     | `#xFF`                | `#xFF`                      |
| sized literal  | -   | `BV.mkBV w 42`  | `(_ bv42 16)`         | `(bv 16 42)`          | `#x…`/`#b…`           | `(bv 42 16)` / `(bv 42 @w)` |
| concat         | -   | `BV.concat`     | `concat`              | `bv-concat`           | `concat`              | `concat`                    |
| extract        | -   | `BV.select`     | `(_ extract hi lo) x` | `bv-select`           | `(_ extract hi lo) x` | `(extract hi lo x)`         |
| zero-extend    | -   | `BV.zext`       | `(_ zero_extend n) x` | `(zero-extend w x)`   | `(_ zero_extend n) x` | `(zext n x)`                |
| sign-extend    | -   | `BV.sext`       | `(_ sign_extend n) x` | `(sign-extend w x)`   | `(_ sign_extend n) x` | `(sext n x)`                |

w4qq drops the SMT/serialize `(_ … )` indexed-identifier wrapper: `extract`,
`zext`, and `sext` take their index arguments directly. Note that w4qq's `bv`
literal is `(bv value width)`, whereas `crucible-syntax` writes it width-first,
`(bv width value)`.

## Metavariables

| Sigil    | Splices                | Example        |
|----------|------------------------|----------------|
| `$name`  | a `SymExpr sym tp`      | `(add $x $y)`   |
| `@name`  | a `NatRepr w` (widths)  | `(bv 42 @w)`    |

## Binding

`(let ((v e) …) body)` - parallel, SMT-style.

## Example

Here's what that same test could look like with quasiquoters based on the above scheme:

```haskell
testRotate1' :: TestTree
testRotate1' = testCase "rotate test1" $ withOnlineZ3 $ \sym s -> do
  bv <- freshConstant sym (userSymbol' "bv") (BaseBVRepr (knownNat @32))

  bv2 <- [w4 sym| (rol (rol $bv (bv 8 32)) (bv 16 32)) |]
  bv3 <- [w4 sym| (rol $bv2 (bv 8 32))  |]
  bv4 <- [w4 sym| (ror $bv2 (bv 24 32)) |]
  bv5 <- [w4 sym| (ror $bv2 (bv 28 32)) |]

  res  <- [checkSat| (ne $bv $bv3) |] sym s
  isUnsat res @? "unsat1"

  res1 <- [checkSat| (ne $bv $bv4) |] sym s
  isUnsat res1 @? "unsat2"

  res2 <- [checkSat| (ne $bv $bv5) |] sym s
  isSat res2 @? "sat"
```

[EDIT]: I wrote all those `bv` args backwards, IMO it should be `(bv WIDTH VALUE)`

langston-barrett:
Fixes #422.

### DIFF
--- a/what4/src/What4/QQ.hs
+++ b/what4/src/What4/QQ.hs
@@ -1,0 +1,622 @@
+-----------------------------------------------------------------------
+-- |
+-- Module           : What4.QQ
+-- Description      : Quasiquoters for What4 symbolic expressions
+-- Copyright        : (c) Galois, Inc 2026
+-- License          : BSD3
+-- Maintainer       : Langston Barrett <langston@galois.com>
+-- Stability        : provisional
+--
+-- This module provides the 'w4' Template Haskell quasiquoter for building What4
+-- symbolic expressions using a small, readable s-expression syntax, extended
+-- with metavariables that splice in-scope Haskell values.
+--
+-- Parsing and code generation happen entirely at compile time: the quasiquoter
+-- parses its body to an s-expression and emits a tree of "What4.Interface"
+-- builder calls.
+--
+-- A quoted term elaborates to a function of the symbolic backend:
+--
+-- > [w4| add $x $y |] :: (IsExprBuilder sym, 1 <= w) => sym -> IO (SymBV sym w)
+--
+-- == Metavariables
+--
+-- [@\$name@]: An /expression/ metavariable. Splices the in-scope Haskell
+--   binding @name@, which must be a @'SymExpr' sym tp@.
+--
+-- [@\@name@]: A /width/ metavariable. Splices the in-scope Haskell binding
+--   @name@, which must be a @'NatRepr' w@. Used in bitvector-literal width
+--   positions, e.g. @(bv \@w 42)@.
+--
+-- == Syntax
+--
+-- Boolean and bitvector operations share @and@, @or@, @xor@, and @not@.
+-- Arithmetic uses @add@, @sub@, @neg@, @mul@, and @abs@; bitvector literals
+-- are @(bv width value)@, @#b0101@, or @#xFF@. Signedness-sensitive
+-- bitvector operations have explicit names such as @ult@ and @sdiv@.
+-- Numeric operators do not implicitly coerce between Int and Real; see
+-- "What4.QQ.Class".
+--
+-- The outermost application may omit its parentheses: @[w4| add $x $y |]@
+-- is equivalent to @[w4| (add $x $y) |]@. Nested applications must remain
+-- parenthesized.
+--
+-- == Operations
+--
+-- === Boolean
+--
+-- +----------------+------------+-----------------+-----------+
+-- | Operation      | SMT-LIB    | What4.Serialize | w4        |
+-- +================+============+=================+===========+
+-- | and            | @and@      | @andp@          | @and@     |
+-- +----------------+------------+-----------------+-----------+
+-- | or             | @or@       | @orp@           | @or@      |
+-- +----------------+------------+-----------------+-----------+
+-- | not            | @not@      | @notp@          | @not@     |
+-- +----------------+------------+-----------------+-----------+
+-- | implies        | @=>@       | -               | @implies@ |
+-- +----------------+------------+-----------------+-----------+
+-- | xor            | @xor@      | @xorp@          | @xor@     |
+-- +----------------+------------+-----------------+-----------+
+-- | if-then-else   | @ite@      | @ite@           | @if@      |
+-- +----------------+------------+-----------------+-----------+
+-- | equal          | @=@        | @=@             | @eq@      |
+-- +----------------+------------+-----------------+-----------+
+-- | not equal      | @distinct@ | -               | @ne@      |
+-- +----------------+------------+-----------------+-----------+
+--
+-- @eq@, @ne@, and @if@ work for every What4 base type; GHC checks that their
+-- operands or branches have matching types.
+--
+-- === Integer and real
+--
+-- +------------------+------------+-----------------+-------+
+-- | Operation        | SMT-LIB    | What4.Serialize | w4    |
+-- +==================+============+=================+=======+
+-- | add              | @+@        | @intadd@        | @add@ |
+-- +------------------+------------+-----------------+-------+
+-- | subtract         | @-@        | -               | @sub@ |
+-- +------------------+------------+-----------------+-------+
+-- | negate           | @-@        | -               | @neg@ |
+-- +------------------+------------+-----------------+-------+
+-- | multiply         | @*@        | @intmul@        | @mul@ |
+-- +------------------+------------+-----------------+-------+
+-- | divide           | @div@/@/@  | @intdiv@        | @div@ |
+-- +------------------+------------+-----------------+-------+
+-- | modulus (Int)    | @mod@      | @intmod@        | @mod@ |
+-- +------------------+------------+-----------------+-------+
+-- | absolute value   | @abs@      | @intabs@        | @abs@ |
+-- +------------------+------------+-----------------+-------+
+-- | less than        | @<@        | -               | @lt@  |
+-- +------------------+------------+-----------------+-------+
+-- | less or equal    | @<=@       | @intle@         | @le@  |
+-- +------------------+------------+-----------------+-------+
+-- | greater than     | @>@        | -               | @gt@  |
+-- +------------------+------------+-----------------+-------+
+-- | greater or equal | @>=@       | -               | @ge@  |
+-- +------------------+------------+-----------------+-------+
+--
+-- @add@, @sub@, @neg@, and @mul@ also work on bitvectors. @div@ is overloaded
+-- between Integer and Real; @mod@ is Integer-only. There are no implicit
+-- Integer/Real coercions.
+--
+-- === Bitvector arithmetic and comparison
+--
+-- +---------------------+----------+-----------------+--------+
+-- | Operation           | SMT-LIB  | What4.Serialize | w4     |
+-- +=====================+==========+=================+========+
+-- | unsigned divide     | @bvudiv@ | @bvudiv@        | @udiv@ |
+-- +---------------------+----------+-----------------+--------+
+-- | signed divide       | @bvsdiv@ | @bvsdiv@        | @sdiv@ |
+-- +---------------------+----------+-----------------+--------+
+-- | unsigned remainder  | @bvurem@ | @bvurem@        | @urem@ |
+-- +---------------------+----------+-----------------+--------+
+-- | signed remainder    | @bvsrem@ | @bvsrem@        | @srem@ |
+-- +---------------------+----------+-----------------+--------+
+-- | unsigned less than  | @bvult@  | @bvult@         | @ult@  |
+-- +---------------------+----------+-----------------+--------+
+-- | unsigned less/equal | @bvule@  | @bvule@         | @ule@  |
+-- +---------------------+----------+-----------------+--------+
+-- | unsigned greater    | @bvugt@  | @bvugt@         | @ugt@  |
+-- +---------------------+----------+-----------------+--------+
+-- | unsigned greater/eq | @bvuge@  | @bvuge@         | @uge@  |
+-- +---------------------+----------+-----------------+--------+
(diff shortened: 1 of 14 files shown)


langston-barrett:
Hmm, CI failing with
```
error: unsupported tarball input attribute 'lastModified'

       … while updating the lock file of flake 'github:GaloisInc/flakes/1b46c6f3de751a4fab8f5b3e491289979969ebbe'
ABC version
```

RyanGlScott:
It's possible that I inadvertently broke something in https://github.com/GaloisInc/flakes/pull/11, which is the most recent commit that updated the [`flakes`](https://github.com/GaloisInc/flakes) repo. That being said, it's not obvious to me why that would be the case. As far as I can tell, that repo's `flake.lock` file was already using the `lastModified` attribute before that PR...

kquick:
I doubt your change caused this, Ryan.  I am more suspicious of https://github.com/cachix/install-nix-action/releases#release-v31.11.0, which bumped the version of `nix` used by the action from 2.34.8 to 2.35.1, which counts as a major release.  @langston-barrett , if you want to try modifying the `test.yml` file to select `cachix/install-nix-action@v31.10.7` instead of just `...@v31`, we can test this hypothesis.

kquick:
Both your pin and the following experiment disprove the hypothesis:

```
$ nix --version
nix (Nix) 2.28.5
$ nix run github:galoisinc/flakes#cvc5.v1_1 -- --show-config
This is cvc5 version 1.1.0
...
cocoa    : yes
...
$

$ nix run github:nixos/nixpkgs#nixVersions.nix_2_35 -- --version
nix (Nix) 2.35.1
$ nix run github:nixos/nixpkgs#nixVersions.nix_2_35 run github:galoisinc/flakes#cvc5.v1_1 -- -- --show-config
This is cvc5 version 1.1.0
...
cocoa    : yes
...
$
```

The above uses my locally installed nix version (2.28) and runs the latest Galois flakes to get the CVC5 configuration, which confirms the availability of cocoa (which was the latest change made to flakes).
It then uses nix to fetch and run nix at version 2.35 and that runs the latest Galois  flakes do the same CVC5 query.

Neither seems to have any problems.
Hrm, indeed.

langston-barrett:
Okay, so it looks like some combination of downgrading and unpinning `nix_path` and `install_url` worked... Thoughts on what we should do? We might *not* want to actually unpin those, but might need to pin them to more recent versions or something? Any thoughts @kquick? (Maybe we could do a fix in a separate PR and I can rebase)

kquick:
My general disposition would be to unpin (and change the install-nix action version back to just `v3`). 

The problem here was caused by pinning to a very old version while updates were committed using newer versions; the nix tooling seems to be reasonably good at migration so I don't anticipate that pinning will have any benefits to outweigh the downside of a future recurrence of what happened here.  It looks like the previous pinning was a result of some attempts to change action versions, which I believe dependabot is now handling.

langston-barrett:
I don't have time to finish this up. I think it would be really nice if we had a general typed-ast (GADT) S-expression library with quasiquoting and de/serialization, but I definitely don't have time for that.

PR #433, opened 2026-08-07
`baseIsConcrete`: Fix `BaseFloatRepr` case

### LINKED ISSUE #432: Incorrect `BaseFloatRepr` case in `baseIsConcrete`
The implementation of `baseIsConcrete` incorrect claims that floats are never concrete:

https://github.com/GaloisInc/what4/blame/905ff37a92b193a142f833b85429b50b5510bb46/what4/src/What4/Interface.hs#L3139

This should be something like `isJust $ asFloat x` instead, similarly to how the other cases are handled.

RyanGlScott:
The previous implementation was blatantly wrong.

Fixes #432.

### DIFF
--- a/what4/CHANGES.md
+++ b/what4/CHANGES.md
@@ -33,6 +33,8 @@
   with a zero denominator.
 * Fix a bug in which calling `solver_adapter_write_smt2 yicesAdapter` would
   always fail.
+* Fix a bug in which `baseIsConcrete` would incorrectly claim that concrete
+  `SymFloat`s are not concrete.
 
 # 1.7.3 -- 2026-01-26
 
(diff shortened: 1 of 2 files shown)


kquick (approved):

kquick:
Clearly we had no tests for this particular behavior.  I don't know that a trivial synthetic test is worth the effort; if you think this will be sufficiently exercised as part of the related crucible work, it's probably not worth adding one, otherwise we should probably leave some trail behind to add a feasible test at some point.

RyanGlScott:
I didn't actually hit this bug in practice---this was something that I happened to notice when inspecting adjacent code. I figured I may as well tidy this up while I was in the neighborhood.

PR #437, opened 2026-09-01
Prepare release 1.8

aschwerdfeger-galois:


RyanGlScott:
The CI is failing on jobs with sufficiently old `cabal-install` versions (i.e., older than `cabal-install-3.10.2.0`) due to https://github.com/haskell/cabal/issues/11111. Ideally, we'd be able to upgrade to newer `cabal-install` versions, although the `what4` CI currently installs `cabal-install` from particular `nixpkgs` versions, and it's unclear to me what would happen if we just bumped the `nixpkgs` version everywhere.

kquick:
I'm reading more about the root of this issue (it's unclear why this suddenly started happening).  We could try a simple fix: https://github.com/BraneFramework/brane/commit/8dfaabff6c1ffe96f03aae5242ebbd71ae31447a in the interim.

kquick:
OK, I believe I understand the issue better now:
* haskell.org rotated the keys
* newer versions of cabal recognize the new keys, but older versions won't recognize the new key
* this CI uses nixpkgs to get a point-in-time set of tools (including older cabal), but since it's not a full flake it does not specify the full set of dependencies, thus cabal still refers to hackage to get the necessary Haskell dependencies.  The same sort of issue would occur for any distribution using an older cabal version.

I can update the CI to use a recent `cabal` but still use the older `ghc`.

RyanGlScott (approved):

